PR #8, opened 2022-03-21
Add tests for new_with_root

ngundotra:
See tests for `new_with_root`

Note: adding/removing the same leaves within the same block will fail because the 2nd instruction will have wrong assumption about preexisting node value

Note: replacing the same node 2x within the same block will cause runtime error

### DIFF
--- a/ConcurrentMerkle-cost.ipynb
+++ b/ConcurrentMerkle-cost.ipynb
@@ -1,0 +1,395 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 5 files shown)


ngundotra:
Added [de475b1](https://github.com/jarry-xiao/merkle-wallet/pull/8/commits/de475b1e9a96761bc7bb75e1ce97a03a1b94e0e6) so that on-chain tree can handle replaces without panicking, and added test verifying expected behavior

PR #16, opened 2022-03-28
Add continuous testing

ngundotra:
Need to start benchmarking performance to figure out what the default `MAX_BUFFER_SIZE` should be.

Key item: when we are processing super old logs -> this can dramatically increase execution cost -> reduce probability of Tx being included in the block.


ngundotra:
Note, can only run 1 gummy roll test at a time

PR #19, opened 2022-03-29
A high level program that can be used to append/transfer/remove a message

steveluscher:
This PR adds an on-chain program that can be used to:

* **append** a proof of inclusion for a given (owner, message) pair
* **transfer** an existing proof of inclusion for a given message to a new owner
* **remove** a proof of inclusion from the tree

The message can be an arbitrarily large vector of bytes.

To **transfer** or **remove**, the original owner must sign the transaction.

There exists protection from modifying the message during a transfer, or removing a proof without knowledge of the original message.

### DIFF
--- a/programs/gummyroll_crud/src/lib.rs
+++ b/programs/gummyroll_crud/src/lib.rs
@@ -52,14 +52,30 @@ pub mod gummyroll_crud {
         gummyroll::cpi::append(cpi_ctx, leaf)
     }
 
-    pub fn transfer(
-        _ctx: Context<Transfer>,
-        _root: [u8; 32],
-        _message: Vec<u8>,
-        _proof: Vec<[u8; 32]>,
-        _index: u32,
+    pub fn transfer<'info>(
+        ctx: Context<'_, '_, '_, 'info, Transfer<'info>>,
+        root: [u8; 32],
+        message: Vec<u8>,
+        index: u32,
     ) -> Result<()> {
-        Ok(())
+        let gummyroll_program = ctx.accounts.gummyroll_program.to_account_info();
+        let merkle_roll = ctx.accounts.merkle_roll.to_account_info();
+        let owner = ctx.accounts.owner.to_account_info();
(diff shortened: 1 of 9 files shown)


programs/gummyroll_crud/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+        );
+        let leaf = Node::new(get_message_hash(&owner, &message).to_bytes());
+        gummyroll::cpi::append(cpi_ctx, leaf)
+    }
+
+    pub fn transfer(
+        ctx: Context<Transfer>,
+        root: [u8; 32],
+        message: Vec<u8>,
+        proof: Vec<[u8; 32]>,
```
jarry-xiao, line 59: Will need to change this to be a list of `Pubkey`s in `ctx.remaining_accounts`

jarry-xiao:
Program looks great! The only change will be modifying the input to use Accounts rather a Vec for the proof, but will merge after that is done

tests/gummyroll_crud.ts
```diff
@@ -0,0 +1,358 @@
...
+          },
+          signers: [signer ?? feePayerKeypair],
+        }
+      );
+      const tx = new Transaction().add(transferIx);
+      await GummyrollCrud.provider.send(tx, [signer ?? feePayerKeypair], {
+        commitment: "confirmed",
+      });
+    }
+    beforeEach(async () => {
```
ngundotra, line 193: Think we only want `appendMessage` to run once at beginning of entire test that way we test this on devnet without burning devnet sol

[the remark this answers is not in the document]
steveluscher: Ahhh, interesting. That's a crappy problem to have :)

In general, I built these tests so that no one test can possibly clobber another (ie. you can run any test on its own, in any order). That necessarily mean that every test has to set up the world from scratch. If you're running that on `devnet`, that costs money :/

PR #23, opened 2022-03-30
Enable all tests, and filter test runner by files that are *actually* tests

steveluscher:


Anchor.toml
```diff
@@ -27,4 +27,4 @@ cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
 [scripts]
-test = "yarn run ts-mocha -t 1000000 tests/gummyroll.ts"
```
steveluscher, line 30: I don't know if this change was on purpose or if this was accidentally committed when you were testing gummyroll alone.

Anchor.toml
```diff
@@ -27,4 +27,4 @@ cluster = "localnet"
...
+test = "yarn run ts-mocha -t 1000000 tests/**/*-test.ts"
```
steveluscher, line 30: This change makes the `gummyroll_crud` tests (et al.) run again, which will catch cross-program breakages. Also, added a `-test` suffix so that mocha doesn't try to run the non-tests in the `tests/` folder.

PR #26, opened 2022-04-05
Added PDA-based authorization scheme to the CRUD program

steveluscher:
Presume that there is a keypair **BRAND** that manages _appends_ and _deletes_ from the tree.

Presume that each leaf encodes an **OWNER** that's allowed to _transfer_ an asset to another **OWNER**.

In this PR:

* Created an authority PDA owned by the CRUD program. The PDA encodes `(TREE_ADDRESS, BRAND)`.
* Added an ‘initialize empty tree’ instruction to the CRUD program. Creates a tree that only the authority PDA can write to.
* The CRUD program newly requires that BRAND is a signer on the add/remove instructions.
* The CRUD program newly requires that OWNER is a signer on the transfer instruction.

### DIFF
--- a/programs/gummyroll_crud/src/lib.rs
+++ b/programs/gummyroll_crud/src/lib.rs
@@ -9,15 +9,12 @@ pub struct CreateTree<'info> {
     #[account(mut)]
     pub authority: Signer<'info>,
     #[account(
-        init,
         seeds = [
             b"gummyroll-crud-authority-pda",
             merkle_roll.key().as_ref(),
             authority.key().as_ref(),
         ],
         bump,
-        payer = authority,
-        space = 0
     )]
     /// CHECK: This account is neither written to nor read from.
     pub authority_pda: UncheckedAccount<'info>,
--- a/programs/gummyroll_crud/src/lib.rs
+++ b/programs/gummyroll_crud/src/lib.rs
@@ -4,32 +4,92 @@ use gummyroll::{program::Gummyroll, Node};
 declare_id!("Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS");
 
 #[derive(Accounts)]
-pub struct Add<'info> {
+#[instruction(max_depth: u32, max_buffer_size: u32)]
+pub struct CreateTree<'info> {
+    #[account(mut)]
+    pub authority: Signer<'info>,
+    #[account(
+        init,
+        seeds = [
+            b"gummyroll-crud-authority-pda",
+            merkle_roll.key().as_ref(),
+            authority.key().as_ref(),
+        ],
+        bump,
+        payer = authority,
+        space = 0
+    )]
+    /// CHECK: This account is neither written to nor read from.
+    pub authority_pda: UncheckedAccount<'info>,
+    pub gummyroll_program: Program<'info, Gummyroll>,
     #[account(mut)]
     /// CHECK: unsafe
     pub merkle_roll: UncheckedAccount<'info>,
-    pub owner: Signer<'info>,
+    pub system_program: Program<'info, System>,
+}
+
+#[derive(Accounts)]
+pub struct Add<'info> {
+    #[account(mut)]
+    pub authority: Signer<'info>,
+    #[account(
+        seeds = [
+            b"gummyroll-crud-authority-pda",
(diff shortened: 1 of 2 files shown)


programs/gummyroll_crud/src/lib.rs
```diff
@@ -4,51 +4,149 @@ use gummyroll::{program::Gummyroll, Node};
...
+        let authority = ctx.accounts.authority.to_account_info();
+        let authority_pda = ctx.accounts.authority_pda.to_account_info();
+        let authority_pda_bump_seed = &[*ctx.bumps.get("authority_pda").unwrap()];
+        let seeds = &[
+            b"gummyroll-crud-authority-pda",
+            merkle_roll.key.as_ref(),
+            authority.key.as_ref(),
+            authority_pda_bump_seed,
+        ];
+        let authority_pda_signer = &[&seeds[..]];
```
steveluscher, line 116: I hate all this duplication – here and in the Anchor `seeds` macro – but my Rust is not strong enough to figure out how to DRY this up. Thoughts?

programs/gummyroll_crud/src/lib.rs
```diff
@@ -4,51 +4,149 @@ use gummyroll::{program::Gummyroll, Node};
...
     pub gummyroll_program: Program<'info, Gummyroll>,
+    #[account(mut)]
+    /// CHECK: unsafe
+    pub merkle_roll: UncheckedAccount<'info>,
 }
 
 #[derive(Accounts)]
 pub struct Remove<'info> {
+    #[account(mut)]
+    pub authority: Signer<'info>,
```
steveluscher, line 54: I remember that you suggested the asset owner should be able to delete leaves, @jarry-xiao, but I made it so that only the tree owner can do so with the plain vanilla add/remove instructions. I thought it would be useful to just have some super-user power to add and remove things from the tree, if it's your tree.

When we have a good use case for an owner having that power (eg. a decompress operation) then maybe we can just add a new instruction?

programs/gummyroll_crud/src/lib.rs
```diff
@@ -4,51 +4,149 @@ use gummyroll::{program::Gummyroll, Node};
...
+        init,
+        seeds = [
+            b"gummyroll-crud-authority-pda",
+            merkle_roll.key().as_ref(),
+            authority.key().as_ref(),
+        ],
+        bump,
+        payer = authority,
+        space = 0
+    )]
```
jarry-xiao, line 21: You don't need to initialize this account :)

programs/gummyroll_crud/src/lib.rs
jarry-xiao: I think this totally depends on context, but we can probably use this same instruction and just check that the authority is either the owner or the super authority

programs/gummyroll_crud/src/lib.rs
jarry-xiao: Yeah this triggers my mild OCD as well, but I've become a believer in "all code that works is good code." This is what working at a trading firm does to you...

programs/gummyroll_crud/src/lib.rs
```diff
@@ -4,51 +4,149 @@ use gummyroll::{program::Gummyroll, Node};
...
-        let owner = ctx.accounts.owner.to_account_info();
-        let cpi_ctx = CpiContext::new(
+        let authority_pda_bump_seed = &[*ctx.bumps.get("authority_pda").unwrap()];
+        let seeds = &[
+            b"gummyroll-crud-authority-pda",
+            merkle_roll.key.as_ref(),
+            authority.key.as_ref(),
+            authority_pda_bump_seed,
+        ];
+        let authority_pda_signer = &[&seeds[..]];
```
jarry-xiao, line 140: If you want, you can just inline this in the `CpiContext::new_with_signer` call

```
        let cpi_ctx = CpiContext::new_with_signer(
             gummyroll_program,
             gummyroll::cpi::accounts::Modify {
                 authority: authority_pda.clone(),
                 merkle_roll,
             },
             &[&seeds[..]],
         )
```

steveluscher:
Thanks for the comments! Smashing this in now without a re-review, but let me know if you find anything more to address and I'll follow it up!

PR #29, opened 2022-04-07
Fix Gummyroll Program Tests

ngundotra:
They can now be run simultaneously.

More work has to be done to check boundary conditions, since we can now support more than MAX_BUFFER_SIZE replacement instructions in a single batch.


jarry-xiao (approved):

tests/continuous_gummyroll-test.ts
```diff
@@ -1,189 +0,0 @@
-import * as anchor from "@project-serum/anchor";
```
jarry-xiao, line 1: Why rename this file?

programs/gummyroll/src/lib.rs
```diff
@@ -780,6 +780,7 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
             self.increment_active_index();
             Some(self.apply_changes(new_leaf, proof, index))
         } else {
+            msg!("FAILING!");
```
jarry-xiao, line 783: Maybe have a slightly more descriptive log?

tests/continuous_gummyroll-test.ts
ngundotra: So that way it gets picked up in `Anchor.toml`

programs/gummyroll/src/lib.rs
ngundotra: Yes will do

ngundotra:
Tests are currently breaking because event listener is not working properly. 

Going to remove the event listener *entirely* from these tests, get them to pass, and then add @steveluscher 's EventParser logic instead

steveluscher:
> …then add @steveluscher 's EventParser logic instead

Yeah, that was a fun commit last night. Soon, I'll send a PR to Anchor so that you can specify the commitment level right in…

```
program.addEventListener('EventName', cb, {commitment: 'confirmed'});
```

Anchor.toml
```diff
@@ -27,4 +27,4 @@ cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
 [scripts]
-test = "yarn run ts-mocha -t 1000000 tests/**/*-test.ts"
+test = "yarn run ts-mocha -t 1000000 tests/**/*-test.ts"
```
steveluscher, line 30: Configure your editor to add trailing newlines so that we don't get these non-changes in PRs!

https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

tests/continuous_gummyroll-test.ts
steveluscher: Wait…

1. Doesn't this make it _not_ get picked up by `anchor test`? The pattern matches anything with `-test.ts` at the end.
2. If you really did intend to rename this file, make sure to use `git mv`. Git move keeps the blame history intact _through renames_. The way that you've done it here, you can no longer look up who wrote what, because the new file is 100% authored (and created) by your user.

ngundotra:
Fixed git history & tests, so that they work again... 

Going forward, need a test for `insertOrAppend`, but that test requires a valid listener, so we can figure out which index leaf was spam-added to.

Test screenshots below
image
image

PR #30, opened 2022-04-08
Add variable length proofs

ngundotra:
For a depth 20 tree, we don't need to submit a length 20 proof (20 extra public keys !!!)

Since the on-chain tree knows its own height (thanks to the `rightmost_proof`), we only need to submit proofs for the height of the subtree. So if tree is height 1, we need 1 key (the sibling)... same logic applies upwards.

This should make interfacing with the tree a little bit faster when its smaller, since easier to serialize & spam


ngundotra:
how do I clean up this git history please just shoot me

ngundotra:
I think I cleaned up the git history

PR #34, opened 2022-04-14
Demo: in-memory tree server

ngundotra:
To run

> cd server
> npm i
> npm run start

then (separate terminal pane)
> anchor test

should work, expect intermittent failures & for it to be very slow... but its starting point

Demo is the following:
Append leaves to the tree in "random order" (actually in sequence) ([1...], ... [64...])

Then query the server to generate proofs that allow us to sort the leaves in descending order
([64... ], [ 63 .... ], ... [1...])

Note: this test is interesting because it brings up some weirdness. We cannot execute an idempotent replace instruction. That is replace leaf A with A... this will fail with no good errors. We also cannot set the leaf hash values to nil whenever we want -- this is special because it is considered a `remove` instruction...

Worth keeping in mind when we clean up & expose this repo


ngundotra:
Drafted this PR so you have to pull this branch to do the demo

ngundotra:
Closed because doesn't scale

PR #36, opened 2022-04-20
🛠 Refactor Plerkle to support multiple plugins 

ngundotra:
Separate out Redis message bus logic, so that it's easier for 3rd parties to develop additional middleware.

Thoughts:
- Tried creating `listeners: Box<Vec<dyn Parser>>` field, but this does not implement `Send` trait... so I skipped thinking about thread safety & moved on to next option
- Define multiple plugins:
- - Gummyroll Plugin: writes Redis data about updating Gummyroll
- - Metadata Plugin: writes Redis data linking leaf hashes to pre-image (metadata)
- - 3rd party Plugin: extra logging

^ to make this really easy, I'm turning Plerkle into a `define_redis_plugin` macro, but we will see how that goes

### DIFF
--- a/plerkle/src/macros.rs
+++ b/plerkle/src/macros.rs
@@ -33,6 +33,8 @@ use crate::programs::gummy_roll::handle_change_log_event;
 
 #[macro_use]
 pub mod macros {
+    use super::*;
+
     macro_rules! define_redis_plugin {
         ($struct:ident, $name:literal, $func:ident) => {
             #[derive(Default)]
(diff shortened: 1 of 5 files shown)


plerkle/src/macros.rs
```diff
@@ -0,0 +1,202 @@
...
+    thiserror::Error,
+};
+use gummyroll_crud::InstructionName;
+use crate::programs::gummy_roll::handle_change_log_event;
+
+#[macro_use]
+pub mod macros {
+    use super::*;
+
+    macro_rules! define_redis_plugin {
```
jarry-xiao, line 42: I think using a trait is preferable to setting up a macro here. Partially a matter of personal preference, but using traits and implementing them is probably easier for the average Rust dev to integrate with

ngundotra:
Indefinitely delaying this PR until plan ironed out

PR #37, opened 2022-04-21
⌨️ Create batchmint CLI 

ngundotra:
To make it easier to upload leaf hashes as CSVs, also added cli tooling to enable GummyrollCrud compatible hashes.


cli/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+        path: outFile,
+        header: ['treeId', 'node_idx', 'seq', 'level', 'hash']
+    });
+
+    log.debug("doing bfs on a tree");
+    const treeIdStr = treeId.toString();
+    const records = bfs(tree, (treeNode, idx) => {
+        return [
+            treeIdStr,
+            idx.toString(),
```
jarry-xiao, line 80: I think this node index isn't calculated correctly. For a tree of height 2, this is the numbering

      1
  2      3
4  5  6  7

It should be easy to convert to this format based on what you currently have

cli/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    const leafHashes = [];
+    leaves = leaves.sort((left, right) => Number(left.leafIndex) - Number(right.leafIndex));
+
+    validateLeaves(leaves);
+
+    leaves.map((leaf) => {
+        leafHashes.push(new PublicKey(leaf.hash).toBuffer());
+    });
+    const numLeaves = 2 ** maxDepth;
+    while (leafHashes.length < numLeaves) {
```
jarry-xiao, line 51: Technically there's no need to include empty leaves in the initial batch mint. This could potentially make the stored tree significantly smaller

jarry-xiao:
This looks really solid as a first pass. Main suggestions:
1. Fix node index
2. Try to not include empty nodes in the CSV (this is lower priority, but could reduce the amount of work needed for the backfill)

cli/utils.ts
ngundotra: In final version, yes will do

cli/utils.ts
ngundotra: Great catch -- yes I was off by one. Didn't realize `node_idx` was 1-indexed

tests/merkle-tree.ts
```diff
@@ -1,6 +1,7 @@
 import { BN } from "@project-serum/anchor";
 import { keccak_256 } from "js-sha3";
 import * as Collections from 'typescript-collections';
+import { getOrCreateAssociatedTokenAccount } from "../deps/solana-program-library/token/js/src";
```
jarry-xiao, line 4: Nit: remove this import

jarry-xiao (approved):

PR #39, opened 2022-04-22
WIP: Bubblegum

jarry-xiao:
This is the initial scaffold of the bubblegum contract. It's still very much a work in progress, but I'm laying out some of the design decisions here, so they can be discussed.

The PR also contains a refactor of gummyroll with 3 main improvements (will need to change the geyser stuff downstream eventually)
1. Adds a sequence number to the change log and the merkle roll struct
2. Creates an `append_authority` that needs to sign (in addition to the regular authority) when creating new leaves
3. Refactors the state into separate files for better code organization

Bubblegum overview:
Each leaf is structured in the following way:
`hash(owner: PK, delegate: PK, nonce: u128, hash(Metadata))`

The nonce is global for the program and is incremented every time the `Mint` instruction is called. The idea is that the decompressed mint account should use the leaf nonce as a PDA seed (deterministic address). Global nonce still allows for CandyMachine style mints to occur against bubblegum, and the resulting NFT's will be minted directly to the tree.

Supports the following operations within the compressed data structure:
1. Mint -> Appends a new leaf to the tree. Must be signed by the mint_authority of the MerkleRoll (opens the possibility for a CandyMachine style contract)
2. Transfer -> Replaces a leaf on the tree. Must be signed by the leaf owner or delegate
3. Burn -> Removes a leaf from the tree. Must by signed by the leaf owner or delegate
4. Redeem -> Removes a leaf from the tree AND creates a voucher with the leaf data. Must by signed by the leaf owner or delegate
5. Decompress (TODO) -> Takes a voucher and expands it into the TokenMetadata + Token + Mint accounts
6. Compress (TODO) -> Reverses Decompress and Redeem. Will call `insert_or_append` to return the data to the Merkle tree


programs/bubblegum/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+};
+use crate::utils::{append_leaf, insert_or_append_leaf, replace_leaf};
+
+declare_id!("BGUMzZr2wWfD2yzrXFEWTK2HbdYhqQCP2EZoPEkZBD6o");
+
+#[derive(Accounts)]
+pub struct InitNonce<'info> {
+    #[account(
+        init,
+        seeds = [b"bubblegum"],
```
ngundotra, line 21: Nit: have as const `NONCE_PREFIX = b"bubblegum"`

programs/bubblegum/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    #[account(
+        seeds = [merkle_roll.key().as_ref()],
+        bump,
+    )]
+    /// CHECK: This account is neither written to nor read from.
+    pub authority: UncheckedAccount<'info>,
+    pub gummyroll_program: Program<'info, Gummyroll>,
+    #[account(zero)]
+    /// CHECK: This account must be all zeros
+    pub merkle_roll: UncheckedAccount<'info>,
```
ngundotra, line 44: Nit: for consistency across ecosystem, might be worth renaming this to `merkle_slab`, or perhaps more descriptively `merkle_bytes`

programs/bubblegum/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// CHECK: This account is chekced in the instruction
+    pub delegate: UncheckedAccount<'info>,
+    #[account(mut)]
+    /// CHECK: unsafe
+    pub merkle_roll: UncheckedAccount<'info>,
+    #[account(
+        init,
+        seeds = [merkle_roll.key().as_ref(), nonce.to_le_bytes().as_ref()],
+        payer = owner,
+        space = 8 + 32 + 32 + 16 + 32 + 4 + 32,
```
ngundotra, line 150: Nit: expose this as const `VOUCHER_SIZE = 8 + 32 + 32 + 16 + 32 + 4 + 32;`

ngundotra (commented):
For someone who doesn't like coding, you really did fantastic job cleaning this code up. Professional quality.

Question: why the global nonce? This seems like it could cause parallelism issues, and would make more sense to be set up as PDA of merkle roll.

Request: we will eventually want to emit `BubbleGum` events as well, so we can pick them up in geyser.

Suggestion: `impl From Context<...> for CpiContext<...Replace>` across `Bubblegum` contexts.[ Anchor examples](https://github.com/project-serum/anchor/blob/282c394666a389ba9e086f974db39de7a6194028/tests/escrow/programs/escrow/src/lib.rs#L175-L189) have this thing where they basically implement `into` for converting a ctx into a cpi ctx. It would be syntatically cool to have this for `redeem`, `burn`, `transfer`, `delegate`, since you already abstracted the cpi info already.

PR #40, opened 2022-04-26
Danenbm/plerkle productization

danenbm:
- Adding accounts and transaction filtering to plerkle.  I used the objects from the Example Postgres Plugin for selecting accounts and transactions.  If we instead want to streamline it for program ID only, I can easily simplify this handling to one program ID selection config item.

- Added initial Messenger trait and moved existing Redis code into an object that implements the Messenger trait.  It's not in any kind of final state, but I wanted to get feedback on the Messenger trait design ideas before going further.

- There's a few spots in the code where I put discussion notes labeled as TODO.


plerkle/src/geyser_plugin_nft.rs
```diff
@@ -1,123 +1,153 @@
...
+            Ok(config) => {
+                // TODO: I used the objects from the Example Postgres Plugin for selecting
+                // accounts and transactions.  I do some additional error handling in case the
+                // config file cannot be read, but otherwise follow the example.  I chose
+                // this because it meets our needs of specifiying program IDs for filtering
+                // (see example-config.json) and having parity with the example may make it
+                // easier for someone to maintain if they already understand the example.
+                // If we instead want to streamline it for program ID only, I can easly simplify
+                // this handling to one program ID selection config item.
+                self.accounts_selector = Some(Self::create_accounts_selector_from_config(&config));
```
austbot, line 133: this is fine I like it makes it flexible at the coset of a little repetition

austbot (approved):
This looks good overall I think we are at the point we need to select wire format, move messenger into a share lib

plerkle/src/geyser_plugin_nft.rs
danenbm: Ok will leave it thanks!

plerkle/src/messenger.rs
```diff
@@ -0,0 +1,300 @@
...
+    pubkeys!(token, "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb");
+    pubkeys!(a_token, "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL");
+}
+
+pub trait Messenger {
+    fn new() -> Result<Self>
+    where
+        Self: Sized;
+
+    fn send_account(&self, account: &ReplicaAccountInfo) -> Result<()>;
```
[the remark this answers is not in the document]
danenbm: will add in next PR as we develop Messenger trait/separate it out to shared lib.

PR #41, opened 2022-04-27
Moving Messenger trait to shared lib

danenbm:
Also adding methods for slot status and block info.


plerkle/src/geyser_plugin_nft.rs
```diff
@@ -248,12 +240,20 @@ impl<T: 'static + Messenger + Default + Send + Sync> GeyserPlugin for Plerkle<T>
...
     ) -> solana_geyser_plugin_interface::geyser_plugin_interface::Result<()> {
         match blockinfo {
-            ReplicaBlockInfoVersions::V0_0_1(block) => {
-                info!("Updating block: {:?}", block);
-                // block.slot
+            ReplicaBlockInfoVersions::V0_0_1(block_info) => {
+                info!("Updating block: {:?}", block_info);
+
+                // Send block info over messenger.
+                match &mut self.messenger {
```
jarry-xiao, line 247: Do you think it's worth considering supporting multiple messengers? Understandable if not critical ATM.

austbot: I think it the future yes, but not sure if we need it now do you ?

danenbm: Super easy to add `for messenger in self.messengers.iter()` now, either way is fine for me to add now or later.

danenbm: I'm new to team so you two have more vision for the product at the moment.  Even though its an easy ask I'll defer to your direction.

PR #43, opened 2022-05-02
Migrate to Anchor 0.24.2

ngundotra:
Requires compiling separate version of anchor in personal github that explicitly requires 1.10.10

Be sure to upgrade submodules after this PR

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "deps/solana"]
 	path = deps/solana
 	url = https://github.com/solana-labs/solana.git
+[submodule "deps/anchor"]
+	path = deps/anchor
+	url = https://github.com/ngundotra/anchor
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -1,6 +1,12 @@
 # Exit on failure
 set -e
 
+# Build on-chain programs first
+anchor build
+cp programs/target/deploy/gummyroll.so docker-vol/gummyroll.so
+cp programs/target/deploy/gummyroll_crud.so docker-vol/gummyroll_crud.so
+cp programs/target/deploy/bubblegum.so docker-vol/bubblegum.so
+
 # Off Chain Setup
 # Speed up by using cargo remote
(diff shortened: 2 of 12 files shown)


.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "deps/solana"]
 	path = deps/solana
 	url = https://github.com/solana-labs/solana.git
+[submodule "deps/anchor"]
+	path = deps/anchor
+	url = https://github.com/ngundotra/anchor
```
jarry-xiao, line 15: 🤔

ngundotra: Screen Shot 2022-05-02 at 12 05 30 PM

ngundotra: I think this is only viable solution

run-docker.sh
```diff
@@ -1,17 +1,22 @@
...
+
+# Build on-chain programs first
+anchor build
+cp programs/target/deploy/gummyroll.so docker-vol/gummyroll.so
+cp programs/target/deploy/gummyroll_crud.so docker-vol/gummyroll_crud.so
+cp programs/target/deploy/bubblegum.so docker-vol/bubblegum.so
+
 # Off Chain Setup
- # Speed up by using cargo remote
-cross build --target x86_64-unknown-linux-gnu --package nft_api
```
danenbm, line 3: I think this will break people trying to build on Mac, because even with correct target, I think cargo build will not work because some dependencies need to compile with x86_64-unknown-linux version of gcc (and not clang).

I am fine with this change because I encountered major problems when trying to get the composed docker container to run properly on M1 Mac, and I doubt anyone else wants to debug that on M1 Mac either.  I switched to building and running this on Linux.

Just mentioning this as an FYI in case Mac compatibility is going to be a requirement for this project.

danenbm (approved):

run-docker.sh
ngundotra: I think we are fine without Mac compatibility. This should be running in cloud anyways

PR #44, opened 2022-05-02
Debug Proof Serving

ngundotra:
Fixes:

- tree/:tree_id not working
- proof/:tree_id/:index


ngundotra:
Should probably have a distinct failure case for when a particular asset cannot be found in the database... as of this commit, we are just returning an empty leaf hash (which is probably correct for most use-cases, but probably not for returning in an endpoint like assets/:tree_id/:index if use-case is seeing flatbuffer'd metadata)

PR #50, opened 2022-05-09
[Server] Only return consistent proofs

ngundotra:
Add recursive query to force max sequence number of nodes in proof to be <= max sequence number of root


ngundotra:
Michael confirmed that this fixed issue (was happening consistently with his client before)

PR #51, opened 2022-05-11
Adding PlerkleSerialization trait to make Messenger trait more generic

danenbm:
## Summary
Beyond making Messenger more generic, the goal was to add some type safety.  Now `Messenger::send()` has to take a type that we expect to be the output of a specific serialization method, rather than a raw byte slice.

Also with this PR did some other minor cleanup of unused code.

After this PR I plan to re-add the `Messenger` receive method.  I have some ideas how to do that with ingest but they are not included yet with this current PR.

## Longer term ideas
A longer term, cleaner option might be to make the `PlerkleSerialization` trait include a `serialize()` method, and then `Messenger` could just take a type and call the type's `serialize()` method before sending it.  The current problem with this approach is that we don't just want to serialize a single type.  For example, the Geyser method called for account updates is:
```
    fn update_account(
        &mut self,
        account: ReplicaAccountInfoVersions,
        slot: u64,
        is_startup: bool,
    ) -> solana_geyser_plugin_interface::geyser_plugin_interface::Result<()> {
```
But I don't want to necessarily serialize everything in `ReplicaAccountInfoVersions`, and I may want `slot` and `is_startup` as well.  So if I wanted to just implement a generic `serialize()` method on a type, I need to create a new type that takes the fields we want.  This is why I did not go further with this serialization trait for now.

## Testing
Continuous gummyroll test still passing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1834,6 +1834,7 @@ dependencies = [
 name = "messenger"
 version = "0.1.0"
 dependencies = [
+ "plerkle-serialization",
  "solana-geyser-plugin-interface",
 ]
 
--- a/messenger/Cargo.toml
+++ b/messenger/Cargo.toml
@@ -9,6 +9,7 @@ documentation = "https://docs.rs/solana-geyser-plugin-interface"
 
 [dependencies]
 solana-geyser-plugin-interface  = { version = "=1.10.10" }
+plerkle-serialization={path= "../plerkle_serialization"}
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
--- a/messenger/src/lib.rs
+++ b/messenger/src/lib.rs
@@ -1,41 +1,14 @@
-use solana_geyser_plugin_interface::geyser_plugin_interface::{
-    ReplicaAccountInfo, ReplicaBlockInfo, ReplicaTransactionInfo, Result, SlotStatus,
+use {
+    plerkle_serialization::PlerkleSerialized,
+    solana_geyser_plugin_interface::geyser_plugin_interface::Result,
 };
 
-pub const ACCOUNT_STREAM: &'static str = "ACC";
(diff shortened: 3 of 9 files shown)


ngundotra:
Right now it's not super clear to me what this change is attempting to accomplish. Maybe it will be more clear once the Messenger receive trait is implemented?

Is this just to make it easier to support AccountInfo indexing for Metaplex later on?

I'd appreciate an example on how this simplifies the existing workflow for parsing Tx info.

danenbm:
> Right now it's not super clear to me what this change is attempting to accomplish. Maybe it will be more clear once the Messenger receive trait is implemented?
> 
> Is this just to make it easier to support AccountInfo indexing for Metaplex later on?
> 
> I'd appreciate an example on how this simplifies the existing workflow for parsing Tx info.

The main benefit of this is that messenger.send() will have some slightly better type safety.  The new code here, I can no longer accidentally send a random slice of bytes.  I have to choose to create a SerializedAccount struct to wrap my bytes, so the caller has to be more intentional about what they are sending.
Example:
```
let bytes: SerializedAccount = serialize_account(&mut builder, account, slot, is_startup);
messenger.send(bytes)
```

I was also hoping that when I implement a receive function with messenger, I can use this type to abstract the stream names and Redis details out of ingest/main.rs.  Then the ingestor could just say `messenger.recv::<SerializedAccount>()` and get the bytes from the right account info stream.  The stream keys could be managed inside of the RedisMessenger implementation.

ngundotra:
I'd say hold off on merging this in until necessary for RedisMessenger. For now I think it's more helpful to focus on setting up for bubblegum & generalizing parsing logic as needed.

danenbm:
Actually the trait is really annoying on the receive side.  I got it working with `Messenger`/`RedisMessenger` receive, but I did not like it at all, so I'm cancelling this PR for now.

PR #53, opened 2022-05-13
Bubblegum parse

ngundotra:
Read NFT metadata into postgres


nft_api/src/ingest/bubblegum_indexer.rs
```diff
@@ -0,0 +1,141 @@
...
+            // Retrieve nonce value from the LeafSchemaEvent emitted
+            let leaf_event_result = get_bubblegum_leaf_event(logs);
+            if leaf_event_result.is_err() {
+                println!("Could not find leaf event");
+                return Err(());
+            };
+            let leaf_event = leaf_event_result.unwrap();
+
+            let owner = pubkey_from_fb_table(keys, instruction.accounts[4] as usize);
+            let delegate = pubkey_from_fb_table(keys, instruction.accounts[4] as usize);
```
jarry-xiao, line 85: These are the same on mint right

programs/bubblegum/src/lib.rs
```diff
@@ -265,6 +265,36 @@ pub struct Compress<'info> {
...
+    Decompress,
+}
+pub fn get_instruction_type(full_bytes: &Vec<u8>) -> InstructionName {
+    let disc: [u8; 8] = {
+        let mut disc = [0; 8];
+        disc.copy_from_slice(&full_bytes[..8]);
+        disc
+    };
+    match disc {
+        [51, 57, 225, 47, 182, 146, 137, 166] => InstructionName::Mint,
```
jarry-xiao, line 288: I wonder if there's a better way to do this. There's probably a `discriminator` Trait that's implemented for each of the instructions

jarry-xiao:
Looks good to me!

programs/bubblegum/src/lib.rs
ngundotra: fair point, will check this one more time before committing

nft_api/src/ingest/bubblegum_indexer.rs
ngundotra: Yes they are

ngundotra: Actually, they are not

programs/bubblegum/src/lib.rs
ngundotra: No there isn't a discriminator trait that does this

PR #54, opened 2022-05-13
Fix instruction ordering

ngundotra:
Outer instructions lacking inner ixs were always appended, rather than appearing in original order.

Now fixed.

Tested on separate branch


danenbm (approved):
Thank you for finding and fixing this!

PR #69, opened 2022-05-23
Refactors the program parser trait and clean up some functions,

austbot:
I need to put Gummy roll crud back in place but I want to get this in front of the team before I spin wheels there.

things I did not do that need to be done. 
I will do all these in seperate PRs
Generalize Storage -> ne pr
factor specific program logic out of batch mint functions -> new pr 
add a logger -> new pr 
refactor schemas and sql -> new pr


nft_ingester/src/main.rs
```diff
@@ -1,59 +1,46 @@
...
     plerkle::async_redis_messenger::AsyncRedisMessenger,
     plerkle_serialization::transaction_info_generated::transaction_info::root_as_transaction_info,
-    regex::Regex,
-    solana_sdk::pubkey::Pubkey,
     sqlx::{self, postgres::PgPoolOptions, Pool, Postgres},
 };
+use messenger::ACCOUNT_STREAM;
+use nft_ingester::parsers::{BubblegumHandler, GummyRollHandler, InstructionBundle, ProgramHandler, ProgramHandlerManager};
+use nft_ingester::utils::{order_instructions, parse_logs};
+use plerkle_serialization::account_info_generated::account_info::root_as_account_info;
```
ngundotra, line 14: would like to stick just one type of import, I'm fine with either `use {}` or  a lot of smaller `use plerkle::` etc.

nft_ingester/src/parsers/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+pub struct InstructionBundle<'a, 'b> {
+    pub message_id: i64,
+    pub txn_id: String,
+    pub instruction: CompiledInstruction<'a>,
+    pub keys: Vector<'b, ForwardsUOffset<transaction_info::Pubkey<'b>>>,
+    pub instruction_logs: Vec<&'b str>,
+}
```
ngundotra, line 54: this is really nice

nft_ingester/src/parsers/gummy_roll.rs
```diff
@@ -141,6 +190,74 @@ pub async fn batch_insert_cl_records(
...
+        .bind(&tree_bytes)
+        .bind(&auth_bytes)
+        .bind(&pid)
+        .execute(pool)
+        .await
+        .unwrap();
+}
+
+
+pub async fn insert_csv_metadata(
```
ngundotra, line 231: okay to move to `/deprecated` location

ngundotra (commented):
The ProgramHandler trait & it's related abstractions look great. Looks like it will simplify things going forward. 

Not 100% sure what the account updates will be used for in this context. I'm guessing they're for supporting generalized indexers?

left some comments saying its cool to drop / delete / move the `batch_insert/write` code

plerkle_serialization/src/lib.rs
```diff
@@ -2,3 +2,5 @@ pub mod account_info_generated;
 pub mod block_info_generated;
 pub mod slot_status_info_generated;
 pub mod transaction_info_generated;
+
+//TODO -> Add From and TO pubkey for flatbuffer key
```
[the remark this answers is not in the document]
austbot: This is a thing where I could maybe zero copy allow fb puby to manufest as a puvbkey from solana sdk

ngundotra (approved):
looks good

nft_ingester/src/utils/logs.rs
```diff
@@ -0,0 +1,62 @@
...
+        println!("No events captured!");
+        Err(())
+    } else {
+        Ok(events)
+    }
+}
+
+pub fn parse_logs<'a>(
+    log_messages: Option<Vector<'a, ForwardsUOffset<&str>>>,
+) -> Result<Vec<(Pubkey, Vec<&'a str>)>, ()> {
```
danenbm, line 31: Nice! looks like this was able to convert to references fairly straightforwardly using the lifetime specifiers to bind the input to the output Vec, and not have to change much of the function's logic.  Hopefully it was just add the `'a` and not have too many extraneous issues.

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -1,31 +1,86 @@
+use anchor_client::anchor_lang::prelude::Pubkey;
+use lazy_static::lazy_static;
+use solana_sdk::pubkeys;
+use std::sync::Arc;
```
danenbm, line 4: Nit: would like to standardize on one import style.  I don't really care which one but don't want to have competing styles.

nft_ingester/src/parsers/gummy_roll.rs
```diff
@@ -23,16 +33,61 @@ pub struct CLRecord {
...
-    logs: &Vec<String>,
+    logs: &Vec<&str>,
     pid: i64,
     pool: &Pool<Postgres>,
-) -> Result<(), ()> {
+) -> Result<(), IngesterError> {
     let change_log_event_vec = if let Ok(change_log_event_vec) = filter_events_from_logs(logs) {
         change_log_event_vec
     } else {
         println!("Could find emitted program data");
```
danenbm, line 89: Nit: Could "not" find.

nft_ingester/src/main.rs
```diff
@@ -1,59 +1,41 @@
...
-#[tokio::main]
-async fn main() {
-    // Setup Postgres.
+async fn setup_manager<'c>(mut manager: ProgramHandlerManager<'c>) -> ProgramHandlerManager<'c> {
+    // TODO setup figment gor db configuration
     let pool = PgPoolOptions::new()
         .max_connections(5)
         .connect("postgres://solana:solana@db/solana")
         .await
         .unwrap();
```
danenbm, line 23: So each time a manager is setup there is a new Pool with 5 connections created?  I thought the idea of the sqlx Pool struct was to create it once and then just send copies of the handle around to different tasks.

From the sqlx core docs:
```
/// `Pool` is `Send`, `Sync` and `Clone`. It is intended to be created once at the start of your
/// application/daemon/web server/etc. and then shared with all tasks throughout the process'
/// lifetime. How best to accomplish this depends on your program architecture.
```
Seems like you could still just pass in a copy to service_transaction_stream:
```
tasks.push(service_transaction_stream(pool.clone()).await);
```
and then you could just pass it into the constructor:
```
let mut manager = ProgramHandlerManager::new(pool);
```
Seems simpler than having to use arc for this use case, but I'm not adamant about this, more just curious of the need for arc for this particular object.

nft_ingester/src/parsers/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl<'a> ProgramHandlerManager<'a> {
+    pub fn new() -> Self {
+        ProgramHandlerManager {
+            registered_parsers: HashMap::new(),
+        }
+    }
+
+    pub fn register_parser(&mut self, parser: Box<dyn ProgramHandler + 'a>) {
```
danenbm, line 31: Overall I really like the ability to register parsers that can be configured to respond to account updates and/or instructions.  I think this design will future proof this code really well.  The `ProgramHandlerManager` struct and `ProgramHandler` trait are pretty cool!

nft_ingester/src/main.rs
austbot: tech me what is best I put the poool in an arc and was considering using arc mutex to sghare it with all tasks

danenbm: According to the [docs for pool](https://docs.rs/sqlx/latest/sqlx/struct.Pool.html) and seeing what ppl do online, in the simplest terms we should be able to do something like:
```
#[tokio::main]
async fn main() {
    let pool = PgPoolOptions::new()
        .max_connections(5)
        .connect("postgres://solana:solana@db/solana")
        .await
        .unwrap();

   let mut tasks = vec![];

    for i in 0..10 {
        let pool = pool.clone();

        tasks.push(tokio::spawn(async move {
            service_something(i, &pool).await;
        }));
    }
}
```
So each task just gets a copy of the connection pool.  It doesn't seem like we need arc or a mutex.  When I read [this section of the docs](https://docs.rs/sqlx/latest/sqlx/struct.Pool.html#why-use-a-pool) it seems like just using the Pool object as-is is meant to solve those issues of reference counting and mutual exclusion on any given connection in the pool.

Caveats:
1.  There used to be [a bug](https://github.com/launchbadge/sqlx/issues/622) with some of this behavior but looks like it was fixed (although potential repro 5 days ago, but not sure if usage is correct, it looks slightly different).
2. You might be seeing the future state more than me with registering more parsers and separating account handlers from transaction handlers, etc. so maybe I'm not seeing the architectural reason to have a new pool per `ProgramHandlerManager`.

nft_ingester/src/parsers/gummy_roll.rs
```diff
@@ -23,16 +33,61 @@ pub struct CLRecord {
     hash: String,
 }
 
+pubkeys!(
+    GummyRollProgramID,
+    "BGUMzZr2wWfD2yzrXFEWTK2HbdYhqQCP2EZoPEkZBD6o"
```
danenbm, line 38: I think this should be "GRoLLMza82AiYN7W9S9KCCtCyyPRAQP2ifBy4v4D5RMD"

nft_ingester/src/main.rs
```diff
@@ -70,39 +52,79 @@ async fn main() {
...
+                let _ = p
+                    .handle_account(&account_update)
+                    .map_err(|e| {
+                        println!("Error in instruction handling {:?}", e);
+                        e
+                    })
+                    .map_err(|e| {
+                        println!("Error in instruction handling {:?}", e);
+                        e
+                    });
```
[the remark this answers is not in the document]
austbot: its just for logging for now

nft_ingester/src/parsers/bubblegum.rs
austbot: will fix

nft_ingester/src/main.rs
austbot: okay so we clone, I thought clone would make 5 X n cnnections

nft_ingester/src/parsers/gummy_roll.rs
```diff
@@ -141,6 +190,74 @@ pub async fn batch_insert_cl_records(
     }
 }
 
+
+pub async fn batch_init_service(
+    pool: &Pool<Postgres>,
```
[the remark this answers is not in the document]
austbot: will do in future pr

nft_ingester/src/main.rs
austbot: I think dups all the connextions

Ingest.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+COPY messenger /rust/messenger
+WORKDIR /rust/nft_ingester
+COPY ./nft_ingester/Cargo.toml ./Cargo.toml
+
+RUN cargo build
+
+COPY ./nft_ingester .
+RUN ls -la
+RUN cargo build
+RUN cp -r /rust/nft_ingester/target/$MODE /rust/bin
```
ngundotra, line 21: what does `$MODE` refer to?

nft_ingester/src/main.rs
ngundotra: looks like this was resolved in most recent change

ngundotra (approved):
Looks good to me. Great work!

PR #78, opened 2022-05-25
Update Bubblegum Hash Schema

jarry-xiao:


Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
-    "nft_ingester",
     "programs/bubblegum",
     "programs/gummyroll",
     "programs/gummyroll_crud",
-    "messenger"
 ]
 exclude = [
     "deps",
-    "programs/",
+    "plerkle",
```
austbot, line 9: uh why?

programs/bubblegum/src/state/leaf_schema.rs
```diff
@@ -3,45 +3,81 @@ use gummyroll::state::node::Node;
...
+    }
 }
 
 #[derive(AnchorDeserialize, AnchorSerialize, Clone, Copy, Default, Debug)]
 pub struct LeafSchema {
+    pub version: Version,
     pub owner: Pubkey,
     pub delegate: Pubkey, // Defaults to owner
     pub nonce: u128,
     pub data_hash: [u8; 32],
```
austbot, line 39: data hash should be at the end probably if we ever want to accomplish the curried hash function.

austbot (approved):

PR #84, opened 2022-05-27
Test Gumball Machine

samwise2:
- Add positive case tests for each public instruction in Gumball Machine
- Small bug fixes in gumball machine
- Small refactoring of shared methods in other tests

One limitation of the current `dispense` NFT test is that it does not assert on state change of the underlying merkle tree. I left this out for now because I think it's made a bit easier with the indexing stuff, which I don't have setup locally. But I think this is also testable locally without indexing via helpers in `tests/merkle-tree.ts`. Marked as TODO to add this when desired approach to asserting on that behaviour is known (using indexing vs tracking proofs locally).

This PR will be followed up by at least one more PR that will:
- Add payments to the `dispense` function in Gumball Machine
- Add test cases for dispensing multiple NFTs with Sol and also a token
- Prevent rare NFT snipping by taking advantage for txn atomicity
- (if desired) negative case test suite paths  -- ensuring bad cases are caught


samwise2:
We probably want to squash these commits?

jarry-xiao:
Currently we don't have a policy to squash commits, but it's definitely a good practice. Feel free to squash and merge whenever you think this is ready

jarry-xiao (approved):

PR #85, opened 2022-05-31
Separate datastructure from gummyroll smart contract

### LINKED ISSUE #79: [Structure] Refactor repo
```
indexer/
  plerkle/
  docker/
  nft_api/
  nft_ingester/
  ...
programs/
  ...
  gummyroll/
    src/
      state/
        merkle_roll.rs, refactor this into separate rust package that can be published separately
concurrent_merkle_tree/
  -> this should be where merkle tree functions live, so easier to test & audit
.archive/
  -> make this a dotfile
```

ngundotra:
Purpose:
1 - Add rust tests for the datastructure (concurrent-merkle-tree)
2 - Make gummyroll easier to read
3 - Make it easier to maintain and improve gummyroll

Also, should help identify issues before audit.

Included in changes:
- Lib tests
- Rust merkle tree reference implementation (copied out of `archive/merkle-accumulator/merkle.rs`)

Closes #79


ngundotra:
note to self: fix NFT_API and test other indexing related packages before merging this in

lib/merkle-tree-reference/src/lib.rs
```diff
@@ -0,0 +1,217 @@
+use solana_program::keccak::hashv;
+use std::cell::RefCell;
+use std::collections::VecDeque;
+use std::iter::FromIterator;
+use std::rc::Rc;
+
+pub type Node = [u8; 32];
```
jarry-xiao, line 7: Feels like we could reuse some of the structs in the main lib for the reference

contracts/programs/gummyroll/Cargo.toml
```diff
@@ -18,3 +18,4 @@ default = []
 [dependencies]
 anchor-lang = { path="../../../deps/anchor/lang" }
 bytemuck = "1.8.0"
+concurrent-merkle-tree = { path="../../../lib/concurrent-merkle-tree" }
```
jarry-xiao, line 21: You need to enable the `sol-log` feature to turn on logs

lib/concurrent-merkle-tree/src/log.rs
```diff
@@ -0,0 +1,34 @@
+#[cfg(feature = "sol-log")]
+use solana_program::{log::sol_log_compute_units, msg};
+
+#[cfg(feature = "sol-log")]
+macro_rules! solana_logging {
```
jarry-xiao, line 5: I'm okay with this module not logging, and handling all logs in the contract.

contracts/programs/gummyroll/src/state/merkle_roll.rs
```diff
@@ -1,15 +1,5 @@
-use crate::state::{
```
jarry-xiao, line 1: This file should probably renamed to `merkle_roll_header.rs`

contracts/programs/gummyroll/src/state/change_log.rs
```diff
@@ -1,7 +1,7 @@
 use crate::state::node::Node;
```
jarry-xiao, line 1: I think the state is sufficiently small to all fit into a single file now

lib/concurrent-merkle-tree/src/tests.rs
```diff
@@ -0,0 +1,141 @@
+#[cfg(test)]
```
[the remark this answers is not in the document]
jarry-xiao: bytemuck has this pattern too: https://github.com/Lokathor/bytemuck

jarry-xiao:
Overall I'm a fan of the changes so far. It makes sense to factor out the data structure itself into a lib. Most of the critiques I have are related to formatting/organization, but I think it's a good start (especially good for finalizing the smart contract). Will continue to comment as this gets closer to completion

contracts/programs/gummyroll/src/lib.rs
```diff
@@ -6,14 +6,14 @@ use anchor_lang::{
...
 pub mod utils;
 
-use crate::state::{
-    merkle_roll::{MerkleRoll, MerkleRollHeader},
-    node::Node,
-};
+use crate::error::GummyrollError;
+use crate::state::{change_log::ChangeLogEvent, merkle_roll::MerkleRollHeader, node::Node};
 use crate::utils::ZeroCopy;
+use concurrent_merkle_tree::{merkle_roll::MerkleRoll, state::Node as TreeNode};
```
jarry-xiao, line 16: Do we still need both node objects? It feels pretty confusing

contracts/programs/gummyroll/src/state/change_log.rs
```diff
@@ -25,78 +28,30 @@ pub struct ChangeLogEvent {
...
             .path
             .iter()
             .enumerate()
-            .map(|(lvl, n)| PathNode::new(*n, (1 << (path_len - lvl as u32)) + (self.index >> lvl)))
+            .map(|(lvl, n)| {
+                PathNode::new(
+                    *n,
+                    (1 << (path_len - lvl as u32)) + (changelog.index >> lvl),
+                )
+            })
```
jarry-xiao, line 47: We should add a comment for what the math is doing here and why node_idx is important

contracts/programs/gummyroll/src/lib.rs
ngundotra: Fixed

contracts/programs/gummyroll/src/state/change_log.rs
ngundotra: Addressed in the Gumdrop changes

ngundotra:
Changed `Node` back into a `type` because I think ecosystem composers will be devs.
Tradeoff is that if we have to change `Node` type to be more bytes, it can be a pain.

Perhaps a better method is following the[ solana `Hash` struct... ](https://github.com/solana-labs/solana/blob/2c8a3f39a6bd0030fa8bf3c8683109af31647878/sdk/program/src/hash.rs#L47)
```rust
pub struct Hash(pub(crate) [u8; HASH_BYTES]);
```

jarry-xiao:
I think the `Hash` type is fine as long as the client interface doesn't change (hope it works well with anchor...)

PR #88, opened 2022-06-02
Gumball Machine Dispense Payments

samwise2:
Changes:
- Add payments for dispensing NFTs (separate instruction for SOL and SPL tokens, reason explained below)
- Add tests for payments, including test that mints multiple NFTs
- Small refactor of utility test methods

**Important Design Decision -- Multiple Dispense Instructions for Native SOL vs SPL Token**

Originally, there was a single dispense function that was to handle projects that wanted to be paid out in SPL Tokens or in native SOL. The issue with this approach, is that it required the client to pass dummy arguments to the instruction, to satisfy the requirements of the accounts context. I.e. there would need to be two TokenAccounts in the instruction context, if the user was making a purchase in native sol, they would need to supply two junk TokenAccounts for the instruction to go through. This seemed confusing, and not ideal. Instead, we've implemented two instructions, one for purchases in native sol and another for purchases in SPL token. The downside here is that the client needs to know which instruction to use, depending on the project's `GumballMachineHeader`. However, they effectively already had to know this in order to pull the correct payment accounts, anyways. And with this approach we do not need any wasted accounts in the instruction context.

### DIFF
--- a/programs/gumball-machine/src/lib.rs
+++ b/programs/gumball-machine/src/lib.rs
@@ -1,8 +1,16 @@
 use anchor_lang::{
     prelude::*,
-    solana_program::{keccak::hashv, sysvar, sysvar::SysvarId},
+    solana_program::{
+        keccak::hashv, 
+        sysvar, 
+        sysvar::SysvarId, 
+        pubkey::Pubkey, 
+        program::{invoke, invoke_signed},
+        system_instruction,
+    },
 };
-use anchor_spl::token::Mint;
+use anchor_spl::token::{Mint, TokenAccount, Token, Transfer, transfer};
+use spl_token::native_mint;
 use bubblegum::program::Bubblegum;
 use bubblegum::state::metaplex_adapter::UseMethod;
 use bubblegum::state::metaplex_adapter::Uses;
@@ -55,13 +63,59 @@ pub struct UpdateHeaderMetadata<'info> {
 }
 
 #[derive(Accounts)]
-pub struct Dispense<'info> {
+pub struct DispenseSol<'info> {
     /// CHECK: Validation occurs in instruction
     #[account(mut)]
     gumball_machine: AccountInfo<'info>,
 
     #[account(mut)]
     payer: Signer<'info>,
+
+    /// CHECK: Validation occurs in instruction
+    #[account(mut)]
+    receiver: AccountInfo<'info>,
+    system_program: Program<'info, System>,
+
+    #[account(
+        seeds = [gumball_machine.key().as_ref()],
+        bump,
+    )]
+    /// CHECK: PDA is checked on CPI for mint
+    willy_wonka: AccountInfo<'info>,
+    /// CHECK: Address is verified
+    #[account(address = SlotHashes::id())]
+    recent_blockhashes: UncheckedAccount<'info>,
+    /// CHECK: Address is verified
+    #[account(address = sysvar::instructions::id())]
+    instruction_sysvar_account: UncheckedAccount<'info>,
+    /// CHECK: PDA is checked in CPI from Bubblegum to Gummyroll
+    /// This key must sign for all write operations to the NFT Metadata stored in the Merkle slab
+    bubblegum_authority: AccountInfo<'info>,
+    /// CHECK: PDA is checked in Bubblegum
+    #[account(mut)]
+    nonce: AccountInfo<'info>,
+    gummyroll: Program<'info, Gummyroll>,
+    /// CHECK: Validation occurs in Gummyroll
+    #[account(mut)]
+    merkle_slab: AccountInfo<'info>,
+    bubblegum: Program<'info, Bubblegum>,
+}
+
+#[derive(Accounts)]
+pub struct DispenseToken<'info> {
+    /// CHECK: Validation occurs in instruction
+    #[account(mut)]
+    gumball_machine: AccountInfo<'info>,
+
+    payer: Signer<'info>,
+    
+    #[account(mut)]
+    payer_tokens: Account<'info, TokenAccount>,
+
+    #[account(mut)]
+    receiver: Account<'info, TokenAccount>,
+    token_program: Program<'info, Token>,
+
     #[account(
         seeds = [gumball_machine.key().as_ref()],
         bump,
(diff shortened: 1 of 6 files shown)


programs/gumball-machine/src/lib.rs
```diff
@@ -96,10 +150,103 @@ pub struct Destroy<'info> {
...
+    assert!(clock.unix_timestamp > gumball_header.go_live_date);
+    let size = gumball_header.max_items as usize;
+    let index_array_size = std::mem::size_of::<u32>() * size;
+    let config_size = gumball_header.extension_len * size;
+    let line_size = gumball_header.extension_len;
+
+    assert!(config_data.len() == index_array_size + config_size);
+    let (indices_data, config_lines_data) = config_data.split_at_mut(index_array_size);
+
+    // TODO: Validate data
```
ngundotra, line 186: this step should be added as a function + test

programs/gumball-machine/src/lib.rs
```diff
@@ -284,87 +433,90 @@ pub mod gumball_machine {
             None => {}
         }
         match bot_wallet {
+            // TODO(sorend): do we want to apply some validation here?
+            // If this will only collect SOL then this is fine, if not then need to validate that it's a TokenAccount with a mint corresponding to project mint
```
ngundotra, line 437: add validation function + test

test should pass in a non-token account as the wallet & initialization should fail :)

programs/gumball-machine/src/state/mod.rs
```diff
@@ -5,7 +5,7 @@ use std::mem::size_of;
 
 #[derive(AnchorSerialize, AnchorDeserialize, Clone, Copy, Zeroable, Pod)]
 #[repr(C)]
-// Current Size: 352 bytes
+// Current Size: 384 bytes
```
ngundotra, line 8: expose this GUMBALL_MACHINE_HEADER_SIZE as a const if needed

tests/gumball-machine-test.ts
```diff
@@ -27,10 +28,11 @@ import {
   decodeGumballMachine, OnChainGumballMachine
 } from "./gumball-machine-serde";
 import NodeWallet from "@project-serum/anchor/dist/cjs/nodewallet";
-import { getAssociatedTokenAddress } from "../deps/solana-program-library/token/js/src";
+import { getAssociatedTokenAddress, createMint, getOrCreateAssociatedTokenAccount, mintTo, getAccount } from "../deps/solana-program-library/token/js/src";
```
ngundotra, line 31: using the js libs from the solana library this way usually causes the other anchor tests to break

I hope I'm wrong here, but do the other tests pass in `anchor test` when you have this?

tests/gumball-machine-test.ts
```diff
@@ -383,21 +392,65 @@ describe("gumball-machine", () => {
...
+        signers: [payer]
+      }
+    );
+
+    const tx = new Transaction().add(dispenseInstr);
+    await GumballMachine.provider.send(tx, [payer], {
+      commitment: "confirmed",
+    });
+
+    // TODO(sorend): assert that the effects of the mint are as expected
```
ngundotra, line 431: add this

tests/gumball-machine-test.ts
```diff
@@ -445,40 +498,23 @@ describe("gumball-machine", () => {
...
-    let noncePDAKey;
-    const GUMBALL_MACHINE_ACCT_SIZE = 8352;
-    const GUMBALL_MACHINE_ACCT_HEADER_SIZE = 352;
+  describe("Testing gumball-machine", async () => {
+    let baseGumballMachineHeader: InitGumballMachineProps;
+    let creatorAddress: Keypair;
+    let gumballMachineAcctKeypair: Keypair;
+    let merkleRollKeypair: Keypair;
+    let noncePDAKey: PublicKey;
+    const GUMBALL_MACHINE_ACCT_HEADER_SIZE = 384;
```
ngundotra, line 514: this should be exposed as a header in `../sdk/gumball-machine/index.tsx` or at least at the top of this test

tests/gumball-machine-test.ts
```diff
@@ -511,42 +547,208 @@ describe("gumball-machine", () => {
...
+        merkleRollKeypair = Keypair.generate();
+        nftBuyer = Keypair.generate();
+      });
+      it("Can dispense single NFT paid in sol", async () => {
+        let creatorPaymentWallet = Keypair.generate();
+
+        baseGumballMachineHeader = {
+          urlBase: Buffer.from("https://arweave.net/Rmg4pcIv-0FQ7M7X838p2r592Q4NU63Fj7o7XsvBHEEl"),
+          nameBase: Buffer.from("zfgfsxrwieciemyavrpkuqehkmhqmnim"),
+          symbol: Buffer.from("pehjjqmrjpfcnttlierdqkxjueqjqjsf"),
```
ngundotra, line 635: LOL how many characters are these?
Pretty sure that symbol < 16 chars for a normal NFT, so decompressing this probably won't work 😆

ngundotra (changes requested):
Great progress!

Left some surface level feedback. Looking forward to reviewing again when the TODOs have been fleshed out.

tests/gumball-machine-test.ts
jarry-xiao: I think Symbol on the TokenMetadata struct actually has a 32 byte max

jarry-xiao: JK I'm wrong it's a Vec<u8> with max len of 8.

jarry-xiao: yuck 🤮

programs/gumball-machine/src/lib.rs
```diff
@@ -96,10 +150,103 @@ pub struct Destroy<'info> {
...
+            gumball_header.name_base,
+            gumball_header.symbol,
+            gumball_header.seller_fee_basis_points,
+            gumball_header.is_mutable != 0,
+            gumball_header.collection_key,
+            None,
+            gumball_header.creator_address,
+            random_config_index,
+            config_line,
+        );
```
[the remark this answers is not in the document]
samwise2: I really like that, can add in some sort of clean up PR

programs/gumball-machine/src/lib.rs
samwise2: @jarry-xiao and I talked abt this briefly today. I think we can skip validation here because the `bot_wallet` is only going to receive SOL. 

But we might need to validate the `receiver` (the address we payout mint fees to) passed to initialize_gumball_machine. @ngundotra do you have thoughts on this? I think the current options @jarry-xiao talked about are:
a) IF `mint` is not native -> check that the `Pubkey` of the `receiver` is the associated token address for `mint` with owner `authority.key()`.
b) Pass `receiver` as an account in `ctx`, check that we can unpack the `receiver` account into some TokenAccount with mint: `mint`
c) Validate nothing, user's fault if they init broken machine

Referring to this: https://github.com/jarry-xiao/candyland/blob/566cd5ae3ce596f29585c5317f0a55cd976b0f40/programs/gumball-machine/src/lib.rs#L263

programs/gumball-machine/src/state/mod.rs
samwise2: I just had this for reference for testing purposes, I can take this out, and move the comment to `gumball-machine-serde.ts` that might be a better place. Its utility was just to facilitate writing client side code

tests/gumball-machine-test.ts
samwise2: For me locally all tests were passing except for gummyroll continuous, which I think you fixed on another branch :)

tests/gumball-machine-test.ts
samwise2: will do, investigating for cleanup PR

programs/gumball-machine/src/lib.rs
samwise2: Not sure what this was referring too tbh, just refactored this instr body into helper. @jarry-xiao what were we looking to validate here?

tests/gumball-machine-test.ts
samwise2: yeah these are just random 32 char strings. If we want to ensure decompress-ability, let's change this https://github.com/jarry-xiao/candyland/blob/566cd5ae3ce596f29585c5317f0a55cd976b0f40/programs/gumball-machine/src/state/mod.rs#L18 to have max size of 8?

programs/gumball-machine/src/lib.rs
ngundotra: ah I see. yeah seems fine to skip validation. bots should be taxed in SOL

programs/gumball-machine/src/state/mod.rs
ngundotra: It's fine

tests/gumball-machine-test.ts
ngundotra: sg

PR #92, opened 2022-06-03
Concurrent Writes into Merkle Tree

ngundotra:
Bubblegum auth now controls an allow-list of accounts that can append directly into the tree. Hard-coded to 5 possible spots.

This makes it easier to make money deploying trees by renting out append allowlist spots to dapps. 

However, if this feature is rebuilt by composing programs, then this may end up being a waste of on-chain bytes. Some experimentation is needed to determine usefulness.


contracts/programs/bubblegum/src/lib.rs
```diff
@@ -46,17 +49,79 @@ pub struct InitNonce<'info> {
 
 #[derive(Accounts)]
 pub struct CreateTree<'info> {
+    #[account(mut)]
     pub tree_creator: Signer<'info>,
     #[account(
+        init,
         seeds = [merkle_slab.key().as_ref()],
         bump,
+        space=GUMMYROLL_TREE_AUTHORITY_SIZE + 64,
```
jarry-xiao, line 58: why is there a +64, I'm scared

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -46,17 +49,79 @@ pub struct InitNonce<'info> {
...
+        payer=tree_creator,
     )]
     /// CHECK: This account is neither written to nor read from.
```
jarry-xiao, line 61: We can remove this check now

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -46,17 +49,79 @@ pub struct InitNonce<'info> {
...
+
+#[derive(Accounts)]
+pub struct TransferTreeDelegate<'info> {
+    pub tree_delegate: Signer<'info>,
+    /// CHECK: this account is not read from
+    pub new_delegate: AccountInfo<'info>,
+    #[account(
+        seeds=[authority.tree_id.as_ref()],
+        bump,
+        mut,
```
jarry-xiao, line 92: Nit: prefer mut as the first constraint, but not a big deal

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -323,14 +388,51 @@ pub mod bubblegum {
...
+        if (index as usize) >= ctx.accounts.authority.append_allowlist.len() {
+            return err!(BubblegumError::AppendAllowlistIndexOutOfBounds);
+        }
+        ctx.accounts.authority.append_allowlist[index as usize] = ctx.accounts.new_append_authority.key();
+        Ok(())
+    }
+
+    pub fn remove_append_authority(
+        ctx: Context<RemoveAppendAuthority>,
+    ) -> Result<()> {
```
jarry-xiao, line 426: I think the procedure here should be:
1) Locate the index i  of the removed PK
2) Remove it
3) Swap value at i with the last occupied index

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -46,17 +49,79 @@ pub struct InitNonce<'info> {
...
+        seeds=[authority.tree_id.as_ref()],
+        bump,
+        mut, 
+        constraint = (*tree_delegate.key == authority.owner) || (*tree_delegate.key == authority.delegate)
+    )]
+    pub authority: Account<'info, GummyrollTreeAuthority>,
+}
+
+#[derive(Accounts)]
+pub struct RemoveAppendAuthority<'info> {
```
jarry-xiao, line 113: I would prefer a system when an append_authority has a fixed number of appends. Once reached, it just auto removes itself

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -323,14 +388,51 @@ pub mod bubblegum {
...
+    pub fn transfer_tree_delegate(
+        ctx: Context<TransferTreeDelegate>,
+    ) -> Result<()> {
+        ctx.accounts.authority.delegate = ctx.accounts.new_delegate.key();
+        Ok(())
+    }
+
+    pub fn set_append_authority(
+        ctx: Context<SetTreeAppendAuthority>,
+        index: u8,
```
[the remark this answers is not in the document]
ngundotra: Fixed

contracts/programs/bubblegum/src/lib.rs
ngundotra: Fixed

contracts/programs/bubblegum/src/lib.rs
ngundotra: Fixed

contracts/programs/bubblegum/src/lib.rs
ngundotra: Meant +8 for discriminator lol, fixed

contracts/programs/bubblegum/src/lib.rs
ngundotra: Basically: create a linked list in memory, remove, then serialize back ?

jarry-xiao: order doesn't matter in the append_list, and this seems like a good way to keep the object packed + prevent duplication

ngundotra: I just did a normal `remove`. What's the benefit of doing a `swap remove`?

jarry-xiao: keeps the append list dense

contracts/programs/bubblegum/src/lib.rs
ngundotra: Addressed in most recent commit

contracts/programs/bubblegum/src/lib.rs
ngundotra: I'll come back to this later for cleaner version of remove

ngundotra: Added `swap remove`

jarry-xiao:
> is this pr still needed/outdated?

Eventually I think it's probably wise to rethink the authorization scheme that we've established. If we will be reusing the same tree for almost everything (obv depends a lot on what the runtime can sustain with the write lock), it probably makes sense to handle the `append` authorization at a higher level. In that regard I think we should use this PR as a starting point for a potentially much more complicated procedure.

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -329,6 +383,15 @@ pub mod bubblegum {
         max_depth: u32,
         max_buffer_size: u32,
     ) -> Result<()> {
+        ctx.accounts.authority.tree_id = ctx.accounts.merkle_slab.key();
+        ctx.accounts.authority.owner = ctx.accounts.tree_creator.key();
+        ctx.accounts.authority.delegate = ctx.accounts.tree_creator.key();
+        // By default, enable the tree creator to write to their own tree
+        ctx.accounts.authority.append_allowlist[0] = AppendAllowlistEntry {
```
jarry-xiao, line 390: I don't know about this one chief

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -346,6 +409,43 @@ pub mod bubblegum {
...
+        ctx.accounts.authority.delegate = ctx.accounts.new_delegate.key();
+        Ok(())
+    }
+
+    pub fn add_append_authority(
+        ctx: Context<SetTreeAppendAuthority>,
+        num_appends: u64,
+    ) -> Result<()> {
+        let empty_key: [u8; 32] = [0; 32];
+        for (i, node) in ctx.accounts.authority.append_allowlist.iter().enumerate() {
```
jarry-xiao, line 431: It is also prudent to check that this authority is not already in the `append_allowlist`

contracts/programs/bubblegum/src/state/authority.rs
```diff
@@ -0,0 +1,88 @@
...
+#[account]
+pub struct GummyrollTreeAuthority {
+    /// Pubkey of merkle roll that this manages
+    pub tree_id: Pubkey,
+    /// How many NFTs have been minted
+    pub count: u64,
+    /// Always able to transfer owner, delegate, modify append_allowlist
+    pub owner: Pubkey,
+    /// Always able to transfer delegate, modify append_allowlist
+    pub delegate: Pubkey,
```
jarry-xiao, line 15: Delegates should NEVER be able to transfer delegates....

contracts/programs/bubblegum/src/state/authority.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    pub fn remove_append_authority(&mut self, allowlist_pubkey: &Pubkey) -> Result<()> {
+        let mut allowlist = self.append_allowlist.to_vec();
+        let mut entries = allowlist.iter();
+        match entries.position(|&allowlist_entry| allowlist_entry.pubkey == *allowlist_pubkey) {
+            Some(idx_to_remove) => {
+                allowlist.swap_remove(idx_to_remove);
+                allowlist.push(AppendAllowlistEntry::default());
+                self.append_allowlist[..allowlist.len()].copy_from_slice(&allowlist);
```
jarry-xiao, line 73: You should be able to do this to the array directly. Not a huge issue, but IMO it's a lot cleaner to find the position of the authority key, perform an index swap with the last element and clear the last element (this is essentially just Fisher-Yates)

contracts/programs/bubblegum/src/state/authority.rs
```diff
@@ -0,0 +1,88 @@
+use crate::error::BubblegumError;
+use anchor_lang::prelude::*;
+
+pub const GUMMYROLL_TREE_AUTHORITY_SIZE: usize = 304 + 8;
+pub const APPEND_ALLOWLIST_SIZE: usize = 5;
```
jarry-xiao, line 5: I think we need to up this number

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -346,6 +409,43 @@ pub mod bubblegum {
...
+            if node.pubkey == Pubkey::new(&empty_key) {
+                ctx.accounts.authority.append_allowlist[i] = AppendAllowlistEntry {
+                    pubkey:ctx.accounts.new_append_authority.key(), 
+                    num_appends
```
jarry-xiao, line 435: This number needs to decrement on mints

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -37,30 +38,88 @@ pub struct CreateTree<'info> {
...
+        mut,
+        seeds=[authority.tree_id.as_ref()],
+        bump,
+        constraint = (*tree_delegate.key == authority.owner) || (*tree_delegate.key == authority.delegate)
+    )]
+    pub authority: Account<'info, GummyrollTreeAuthority>,
+}
+
+#[derive(Accounts)]
+pub struct SetTreeAppendAuthority<'info> {
```
jarry-xiao, line 85: I think this workflow is pretty tricky. I wonder if it's the right approach to something like:
1) User requests to be an append_authority with a certain size
2) Tree creator needs to approve the request and append_authority gets granted N mints
3) Tree creator can be a governance thing in the future because decentralization or something like that

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -37,30 +38,88 @@ pub struct CreateTree<'info> {
...
+#[derive(Accounts)]
+pub struct TransferTreeDelegate<'info> {
+    pub tree_delegate: Signer<'info>,
+    /// CHECK: this account is not read from
+    pub new_delegate: AccountInfo<'info>,
+    #[account(
+        mut,
+        seeds=[authority.tree_id.as_ref()],
+        bump,
+        constraint = (*tree_delegate.key == authority.owner) || (*tree_delegate.key == authority.delegate)
```
jarry-xiao, line 79: delegate cannot transfer itself

contracts/programs/bubblegum/src/state/authority.rs
```diff
@@ -0,0 +1,88 @@
+use crate::error::BubblegumError;
+use anchor_lang::prelude::*;
+
```
jarry-xiao, line 3: I think there should probably be some way of tracking the number of remaining spots in the tree to make sure that we don't grant append authority carelessly. In practice, append auth is probably going to just be the tree creator (very centralized), but there should be a system to allow for shared trees

contracts/programs/bubblegum/src/lib.rs
jarry-xiao: Tree creator should be like super admin. Should be able to do anything (abstracts away authorization to higher contract like gov if necessary)

jarry-xiao (changes requested):
We should discuss the details carefully spec out how this should look. I think the first step is to just build a PR that removes append_authority from bubblegum, and then carefully think about how to do this. The nice thing is that the 2 are orthogonal

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -346,6 +409,43 @@ pub mod bubblegum {
...
     }
 
+    pub fn transfer_tree_owner(
+        ctx: Context<TransferTreeOwner>,
+    ) -> Result<()> {
+        ctx.accounts.authority.owner = ctx.accounts.new_owner.key();
+        Ok(())
+    }
+
+    pub fn transfer_tree_delegate(
```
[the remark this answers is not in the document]
ngundotra: I think it was meant to be a pubkey with same privileges as tree creator to modify other `appendAuthority`es

contracts/programs/bubblegum/src/state/authority.rs
ngundotra: Honestly this scheme doesn't even make sense

ngundotra:
Closing this PR in favor of separating:
- Gummyroll: appendAuthority <-> authority unification
- Bubblegum: request-based appendAuthority

contracts/programs/bubblegum/src/state/authority.rs
jarry-xiao: yea we can also do it with the PDA system (need more accounts)

contracts/programs/bubblegum/src/lib.rs
jarry-xiao: If this is the case then tree creator should have not the ability to change the delegate

PR #98, opened 2022-06-09
Fix CMT Errors

ngundotra:
Issue with appends after replacing the RMP leaf or RMP index + 1


ngundotra:
props to @jarry-xiao  for fixing error

PR #99, opened 2022-06-09
[Jank indexer] Rest Api for Proofs

ngundotra:
Also: 
* restructure the package to have different imports than the rest of `contracts` for mental sanity
* add scripts to `package.json`

To use:
```bash
yarn validator &
yarn test
# wait a few seconds, then ctrl-c
yarn server
# go to http://localhost:3000/proof?leafHash=<any leaf hash from yarn test output>
```


ngundotra:
Changes incorporated into jankle

PR #100, opened 2022-06-09
Sugar Shack Contract, Test + SDK

samwise2:
- [x] `Sugar Shack` example Marketplace smart contract working with compressed NFTs. (See design doc here: https://www.notion.so/solanalabs/Compressed-Marketplace-Design-Doc-d4a4c6394d884d09a3504a34941fe73f)
- [x] `sugar-shack-test.js` Test suite exercising positive and negative cases for all core instructions: `create_or_modify_listing`, `remove_listing` and `purchase`.
- [x] Solita generated SDK + manual helpers generated for devs to work with the contract and test it out.
- [ ] ~~`Bubblegum` instruction `Transfer` split into two, one for transferring NFTs with the `owner` as signer, and one for transferring NFTs with `delegate` as signer. This should allow easier composability with other anchor programs (i.e. Marketplace contracts).~~


contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+
+declare_id!("9T5Xv2cJRydUBqvdK7rLGuNGqhkA8sU8Yq1rGN7hExNK");
+
+// TODO(sorend): port to a state directory/mod
+#[account]
+#[derive(Copy)]
+pub struct MarketplaceProperties {
+    // The address to which the marketplace will receive royalties
+    pub payment_receiver: Pubkey,
+    // The royalty percentage (not basis points) the marketplace will receive upon purchase of an NFT
```
jarry-xiao, line 30: Let's make this bps, 1+% marketplace fees already disgust me

contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        data_hash: [u8; 32],
+        root: [u8; 32],
+        nonce: u128,
+        index: u32,
+        creator_shares: Vec<u8>
+    ) -> Result<()> {
+        // First, payout the marketplace's royalty fee. This has to be off the top since Bubblegum mint currently
+        // sets the base creator's share to 100%.
+        // TODO(sorend): consider using bigdecimal for casting and intermediate operations to avoid loss of information for projects with high price 
+        let amount_to_pay_marketplace = (price as f64 * (ctx.accounts.marketplace_props.share as f64/100.0)) as u64;
```
jarry-xiao, line 285: Floating point feels like it could get a little nasty. We can check this when we can get tests working

contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        );
+
+        // CPI to Bubblegum to transfer the NFT to its new owner
+        let price_seed = price.to_le_bytes();
+        let program_id_seed = id();
+        let seeds = &[price_seed.as_ref(), program_id_seed.as_ref(), &[*ctx.bumps.get("listing_delegate").unwrap()]];
+        let authority_pda_signer = &[&seeds[..]];
+        let cpi_ctx = CpiContext::new_with_signer(
+            ctx.accounts.bubblegum.to_account_info(),
+            bubblegum::cpi::accounts::Transfer {
```
jarry-xiao, line 343: My gut tells me that anchor CPI is going to bungle this because there's no ability to specify an optional signer

contracts/programs/gummyshop/src/lib.rs
samwise2: LOL sounds good

contracts/programs/gummyshop/src/lib.rs
samwise2: Agree, upgraded this in the latest contract. I think it should give us more safety in general, but might not work for projects with ridiculously cheap prices.

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -99,15 +99,15 @@ pub struct Burn<'info> {
 }
 
 #[derive(Accounts)]
-pub struct Transfer<'info> {
+pub struct TransferWithOwner<'info> {
```
jarry-xiao, line 102: Let's get this working with the Indexer :)

jarry-xiao: Also for the sake of not breaking existing interfaces, it might be better to do this the sketchy way for the time being

jarry-xiao: Or just not rename the normal `Transfer` Instruction

contracts/programs/gummyshop/src/state/mod.rs
```diff
@@ -0,0 +1,14 @@
+use anchor_lang::prelude::*;
+
+#[account]
+#[derive(Copy)]
+pub struct MarketplaceProperties {
+    // The address to which the marketplace will receive royalties
+    pub payment_receiver: Pubkey,
```
jarry-xiao, line 7: Because we're account constrained, I think it could make sense to transfer the funds directly to this account and have the `payment_receiver` collect any lamports over rent exemption afterwards.

contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+                &[
+                    ctx.accounts.purchaser.to_account_info(),
+                    current_creator_info.clone(),
+                    ctx.accounts.system_program.to_account_info()
+                ]
+            )?;
+            amount_paid_out_to_creators += amount_to_pay_creator;
+            creator_data.push([current_creator_info.key().as_ref(), &[share]].concat());
+        }
+        total_remaining_price_allocation -= amount_paid_out_to_creators;
```
jarry-xiao, line 286: It's important to check for overflows here. It's unlikely to happen, but always good to do your due diligence

contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+                ctx.accounts.purchaser.to_account_info(),
+                ctx.accounts.marketplace_receiver.to_account_info(),
+                ctx.accounts.system_program.to_account_info()
+            ]
+        )?;
+
+        // Second, payout each "creator". Creators are an immutable set of secondary marketplace sale royalty recipients.
+        // Simultaneously, collect <address, share> pairs to prepare to compute creator_hash
+        let mut total_remaining_price_allocation = price - amount_to_pay_marketplace;
+        let basis_point_of_remaining_price_allocation = total_remaining_price_allocation/(10000 as u64);
```
jarry-xiao, line 264: Generally look to use safe math when possible

contracts/programs/gummyshop/src/state/mod.rs
samwise2: Cool, will do this

contracts/programs/gummyshop/src/lib.rs
```diff
@@ -0,0 +1,430 @@
...
+        let marketplace_props_data = &mut ctx.accounts.marketplace_props;
+        assert!(royalty_share <= 10000);
+        marketplace_props_data.share = royalty_share;
+        marketplace_props_data.authority = authority;
+        marketplace_props_data.bump = *ctx.bumps.get("marketplace_props").unwrap();
+        Ok(())
+    }
+
+    /// Enables the authority of the marketplace to update admin properties 
+    pub fn update_marketplace_properties<'info>(
```
samwise2, line 209: There aren't currently tests for this instruction because I wasn't sure if it would be desired, but can add if it sticks.

imo benefits:
- authority can be changed (good for key leaks, etc.)
- share can change (good for marketplace flexibility)

imo cons:
- people may list at fee x, and then have fee y taken when the marketplace changes their props. Could be unexpected to the lister. 

For an example implementation it might be worth having, just to show marketplaces the most flexibility, curious about your thoughts @jarry-xiao

contracts/sdk/bubblegum/idl/bubblegum.json
```diff
@@ -893,6 +893,12 @@
           },
           {
             "name": "DecompressV1"
+          },
```
samwise2, line 896: Note: I haven't changed bubblegum, I just re-ran `node soltia.js` with the latest on main

samwise2:
Updates since last review @jarry-xiao : 

- using safe math now
- purchase takes one less account and funds are paid directly to the marketplace singleton PDA
- withdraw function for an "authority" to withdraw fees from the marketplace PDA to any external key
- update properties function to update the props of the marketplace moving forward
- bubblegum::Transfer unchanged, and CPI hacked. I tried to keep as much Anchor stuff as possible but didn't see a way to nicely pull out a mutable `Instruction` that could be passed to invoke_signed. Tried to instantiate each field of `invoke_signed` with slightly modified Anchor stuff tho
- changed tests to work w tree depth 22 and buffer size 2048, purchase seems to work without issues

samwise2:
name ideas: `sugarshack` , `sweet-touth`

samwise2:
merging via offline approval from @jarry-xiao 🎉

PR #102, opened 2022-06-14
Redis Messenger Generalize to Async Trait

danenbm:
### Making Messenger an async trait
- Moved redis messengers to messenger crate.
- Added cargo feature to enable/disable.
- Renaming async Redis messenger to Redis messenger.
- Ingester and Geyser updated.
- General clean up and improving error handling.

### Testing
- Bubblegum test passes.
- Continuous gummyroll test does not pass on main nor this branch at the moment.

### TODO
- ~~Try to not use tokio Mutex.~~ DONE in https://github.com/jarry-xiao/candyland/pull/104


messenger/src/redis_messenger.rs
```diff
@@ -77,7 +83,8 @@ impl Messenger for RedisMessenger {
         }
     }
 
-    fn send(&mut self, stream_key: &'static str, bytes: &[u8]) -> Result<()> {
+    //impl Future<Output = Result<()>> + 'a {
+    async fn send(&mut self, stream_key: &'static str, bytes: &[u8]) -> Result<()> {
         // Check if stream is configured.
```
sheldon-metaplex, line 88: When streaming time series data, we need to think carefully about the order of messages coming through each step of the pipeline. If messages are guaranteed to be delivered in order of time, the write operations in the database may have significant performance improvements. This may or may not be the case for our current data store.  But, if things get out of order, then the database may have to re-order the datapoints itself, which could be problematic. 
Again this is theoretical, but here's an example. Async message publication to Redis results in out-of-order time series data in the stream. The consumer app consumes the messages sequentially from the stream and publishes write operations in small batches. If the data is unordered, the database may have to sort before writing to disk. 
I'm not sure that anything is problematic in this case, but this is the kind of thing we should be thinking about.

messenger/src/redis_messenger.rs
```diff
@@ -95,56 +102,51 @@ impl Messenger for RedisMessenger {
...
-        Ok(())
-    }
+            Ok(reply) => reply,
+            Err(e) => {
+                error!("Redis receive error: {e}");
+                return Err(GeyserPluginError::Custom(Box::new(
+                    MessengerError::ReceiveError { msg: e.to_string() },
+                )));
+            }
+        };
```
sheldon-metaplex, line 145: What happens to messages that fail to process? This can vary dramatically by system  and configuration, and I'll have to read up on Redis Streams.

nft_ingester/src/main.rs
```diff
@@ -38,7 +39,7 @@ async fn main() {
         .await
```
sheldon-metaplex, line 39: We ought to move the Postgres connection params in line 39 into a configuration file so we can use unique Postgres databases in dev, staging, production.

sheldon-metaplex (commented):
Let's discuss tomorrow, very exciting!

ngundotra:
Looks good 👍

messenger/src/error.rs
```diff
@@ -0,0 +1,13 @@
...
+pub enum MessengerError {
+    #[error("Error creating connection: ({msg})")]
+    ConnectionError { msg: String },
+
+    #[error("Error sending data: ({msg})")]
+    SendError { msg: String },
+
+    #[error("Error receiving data: ({msg})")]
+    ReceiveError { msg: String },
+}
```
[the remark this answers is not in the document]
danenbm: Yes definitely I envision something like this for the long term:
```
if let  Err(e) = sender.send(data).await {
    publish_metric("could not send account info", e, 1);
}
```

messenger/src/redis_messenger.rs
danenbm: Talked to @austbot and you offline about this.  I think we will plan to use a sequence number on the consumer side to deal with this for this project, and it will be fine for this specific case, but I definitely see the more generic use case being applicable.

messenger/src/redis_messenger.rs
danenbm: Right now we will ignore these errors and just the tests will fail.  Created issue to track this design work:
https://app.zenhub.com/workspaces/protocol-engineering-61eebb497fb4920012651e19/issues/metaplex-foundation/metaplex-program-library/537

nft_ingester/src/main.rs
danenbm: Created issue to track this: https://app.zenhub.com/workspaces/protocol-engineering-61eebb497fb4920012651e19/issues/metaplex-foundation/metaplex-program-library/538

danenbm:
Reviewed offline with @austbot and PR was approved by him. Merging now.

PR #103, opened 2022-06-14
Jxiao/tree nonces

jarry-xiao:


ngundotra:
Looks good! Literally only a two liner change. 

Can you just remove the `contracts/package.json` change? Solita not needed at the top-level, feels cleaner to remove it

PR #114, opened 2022-06-17
Upsert

austbot:
Upsert, atomicity, multiple nonce accounts, and fully implemented schema


programs/bubblegum/src/lib.rs
```diff
@@ -64,31 +65,40 @@ pub struct Mint<'info> {
...
     pub authority: UncheckedAccount<'info>,
     #[account(
-        mut,
-        seeds = [NONCE_PREFIX.as_ref()],
-        bump,
+    mut,
+    seeds = [NONCE_PREFIX.as_ref(), merkle_slab.key().as_ref()],
+    bump,
     )]
     pub nonce: Account<'info, Nonce>,
```
jarry-xiao, line 78: given than we have 1 nonce per tree now, we should definitely consolidate the nonce and authority into the same account -> 2x max tree capacity

programs/bubblegum/src/lib.rs
```diff
@@ -332,10 +344,14 @@ pub mod bubblegum {
...
 
     pub fn mint(ctx: Context<Mint>, version: Version, message: MetadataArgs) -> Result<()> {
+        // TODO -> Pass collection in check collection authority or collection delegate authority signer
+        // TODO -> Separate V1 / V1 into seperate instructions
+        //
         let owner = ctx.accounts.owner.key();
         let delegate = ctx.accounts.delegate.key();
         let merkle_slab = ctx.accounts.merkle_slab.to_account_info();
         let nonce = &mut ctx.accounts.nonce;
+        let asset_id = ctx.accounts.asset_id.key();
```
jarry-xiao, line 354: Why does the asset_id need to be passed in the instruction at all? I think it should be `FPA(tree_id, authority, token_program)`. We can log the hash and store it in the indexer

programs/bubblegum/src/lib.rs
```diff
@@ -385,17 +401,19 @@ pub mod bubblegum {
         // Transfers must be initiated either by the leaf owner or leaf delegate
         assert!(owner.is_signer || delegate.is_signer);
         let new_owner = ctx.accounts.new_owner.key();
-        let previous_leaf = LeafSchema::new(
-            version,
+        let (nonce_account, _) = get_nonce_account_id(merkle_slab.key);
+        let (asset_id, _) = get_asset_id(&nonce_account, nonce);
+        let previous_leaf = LeafSchema::new_v0(
+            asset_id,
```
jarry-xiao, line 407: I don't see an explicit reason for the `asset_id` to be a member of this struct

contracts/sdk/bubblegum/src/generated/types/LeafSchema.ts
```diff
@@ -61,7 +61,7 @@ export const leafSchemaBeet = beet.dataEnum<LeafSchemaRecord>([
         ['dataHash', beet.uniformFixedSizeArray(beet.u8, 32)],
         ['creatorHash', beet.uniformFixedSizeArray(beet.u8, 32)],
       ],
-      'LeafSchemaRecord["V0"]'
+      'LeafSchemaRecord["V1"]'
     ),
   ],
-]) as beet.FixedSizeBeet<LeafSchema> | any
+]) as beet.FixedSizeBeet<LeafSchema>
```
jarry-xiao, line 67: This is a solita bug (I think). Without the `| any` hack I get the following error:
```
FixedSizeBeet<EnumDataVariant<"V1", { id: PublicKey; owner: PublicKey; delegate: PublicKey; nonce: bignum; dataHash: number[]; creatorHash: number[]; }>, Partial<...>>; toFixedFromValue(val: any): FixedSizeBeet<...>; description: string; }' to type 'FixedSizeBeet<{ __kind: "V1"; } & { id: PublicKey; owner: PublicKey; delegate: PublicKey; nonce: bignum; dataHash: number[]; creatorHash: number[]; }, Partial<{ __kind: "V1"; } & { ...; }>>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ toFixedFromData(buf: Buffer, offset: number): FixedSizeBeet<EnumDataVariant<"V1", { id: PublicKey; owner: PublicKey; delegate: PublicKey; nonce: bignum; dataHash: number[]; creatorHash: number[]; }>, Partial<...>>; toFixedFromValue(val: any): FixedSizeBeet<...>; description: string; }' is not comparable to type 'ElementCollectionFixedSizeBeet<{ __kind: "V1"; } & { id: PublicKey; owner: PublicKey; delegate: PublicKey; nonce: bignum; dataHash: number[]; creatorHash: number[]; }, Partial<{ __kind: "V1"; } & { ...; }>>'.
  ```
  Didn't want to think about it, but it's probably worth resolving at some point

**Comment by jarry-xiao** [comment]
TS Indexer is currently broken. We should merge as soon as that's fixed

**Comment by jarry-xiao** [comment]
Not in this PR, but a note about `Transfer` and `Burn`. I think to make the IDL a little friendlier, it might make sense to create separate `TransferWithDelegate` and `BurnWithDelegate` instructions because Anchor doesn't let you autogenerate optional signer fields. This separation allows us to have very explicit use cases for each instruction and we can say that `owner` MUST sign for `Transfer` and `Burn` and `delegate` MUST sign for the new ixs.

File: nft_ingester/src/parsers/bubblegum.rs [hunk 5]
```diff
@@ -181,17 +310,19 @@ async fn handle_bubblegum_instruction<'a, 'b>(
         //      otherwise, it becomes hard to reinsert data on a CancelRedeem
         bubblegum::InstructionName::Redeem => {
             println!("Bubblegum: Redeem");
```
jarry-xiao, line 312: You definitely want to save redeemed events in the DB (should function a lot like a Burn)

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -181,17 +310,19 @@ async fn handle_bubblegum_instruction<'a, 'b>(
...
         }
-        bubblegum::InstructionName::CancelRedeem => {
-            println!("Bubblegum: CancelRedeem");
-            // TODO(): nothing
-        }
-        bubblegum::InstructionName::Decompress => {
-            println!("Bubblegum: Decompress");
-            // TODO(): set nonce uuid to a non-queryable state
-        }
-        _ => println!("Bubblegum: Ignored instruction"),
```
jarry-xiao, line 194: TODO implement all of the instructions that map to ReplaceLeaf

jarry-xiao: (i.e. Burn, Delegate, Redeem)

austbot: yes of course

nft_ingester/src/parsers/bubblegum.rs
austbot: yes will do just getting the main stuff done

PR #116, opened 2022-06-19
adds more validation to bgum and parses redeem and cancel redeem

austbot:


contracts/programs/bubblegum/src/state/mod.rs
```diff
@@ -35,9 +35,18 @@ impl Voucher {
...
     pub metadata: MetadataArgs,
     pub nonce: u64,
 }
+
+#[event]
+pub struct NFTDecompressionEvent {
+    pub version: Version,
+    pub id: Pubkey,
+    pub tree_id: Pubkey,
+    pub nonce: u64,
```
jarry-xiao, line 51: Any reason why we need an entirely new event for this?

jarry-xiao: Understandable if it's just to make indexing easier to manage

austbot: its temporary until I have non compressed indexing

PR #117, opened 2022-06-20
api init refactor to real api

austbot:


contracts/sdk/indexer/indexer.ts
```diff
@@ -40,12 +40,13 @@ async function main() {
...
     "confirmed"
   );
+  let maxSeq = 0;
   while (true) {
     try {
       const trees = await db.getTrees();
       for (const [treeId, depth] of trees) {
         console.log("Scanning for gaps");
-        let maxSeq = await fetchAndPlugGaps(connection, db, 0, treeId, {
+        maxSeq = await fetchAndPlugGaps(connection, db, maxSeq, treeId, {
```
ngundotra, line 49: good catch! We should have had this initially

contracts/tests/bubblegum-test-rpc.ts
```diff
@@ -0,0 +1,512 @@
...
+  root: string,
+  proof: AccountMeta[]
+}
+
+async function getProof(asset: PublicKey): Promise<TreeProof> {
+  let resp = await fetch("http://localhost:9090", {
+    method: 'POST',
+    headers: {'Content-Type': 'application/json'},
+    body: JSON.stringify({
+      jsonrpc: "2.0", id: "stupid", method: "get_asset_proof", params: [asset.toBase58()]
```
ngundotra, line 70: This is very useful. Did not know how to structure RPC calls via fetch like this

das_api/src/api_impl.rs
```diff
@@ -0,0 +1,138 @@
...
+        })
+    }
+}
+
+pub fn not_found(asset_id: &String) -> DbErr {
+    DbErr::RecordNotFound(format!("Asset Proof for {} Not Found", asset_id))
+}
+
+#[async_trait]
+impl ApiContract for DasApi {
```
ngundotra, line 44: Every time I see `DasApi` I think I'm reading something Germanic 😆

digital_asset_types/src/rpc/asset.rs
```diff
@@ -0,0 +1,247 @@
...
+    #[serde(rename = "NFT")]
+    Nft,
+    #[serde(rename = "FungibleAsset")]
+    FungibleAsset,
+    #[serde(rename = "Custom")]
+    Custom,
+    #[serde(rename = "Identity")]
+    Identity,
+    #[serde(rename = "Executable")]
+    Executable,
```
ngundotra, line 34: Is this from DAS 2.0?

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -58,6 +61,41 @@ pubkeys!(
...
+impl Display for DownloadMetadata {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "DownloadMetadata from {} for {}", self.uri, self.asset_data_id)
+    }
+}
+
+#[async_trait]
+impl BgTask for DownloadMetadata {
+    async fn task(&self, db: &DatabaseConnection) -> Result<(), IngesterError> {
+        let body: serde_json::Value = reqwest::get(self.uri.clone()) // Need to check for malicious sites ?
```
ngundotra, line 81: pain

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -58,6 +61,41 @@ pubkeys!(
...
+
+impl Display for DownloadMetadata {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "DownloadMetadata from {} for {}", self.uri, self.asset_data_id)
+    }
+}
+
+#[async_trait]
+impl BgTask for DownloadMetadata {
+    async fn task(&self, db: &DatabaseConnection) -> Result<(), IngesterError> {
```
ngundotra, line 80: Is this task also supposed to write to docker disk?

ngundotra (approved):
This PR is super exciting! It's really cool to see the Digital Asset spec move in to place with the Compressed NFTs + existing NFTs interactions.

With these changes, `plerkle` gets much easier to read. Great work. I am in favor of merging this in as-is. I will rebase #121 on top then merge it.

contracts/sdk/indexer/indexer.ts
danenbm: There might be an issue with this because the maxSeq is tree specific, right?  So to keep using maxSeq each time maybe would need to store it in a hashmap [key: treeId, val: maxSeq]

das_api/src/error.rs
```diff
@@ -0,0 +1,24 @@
...
+    DatabaseConnectionError(#[from] sqlx::Error),
+    #[error("Pubkey Validation Err {0} is invalid")]
+    PubkeyValidationError(String),
+    #[error("Validation Error {0}")]
+    ValidationError(String),
+    #[error("Database Error")]
+    DatabaseError(#[from] sea_orm::DbErr),
+}
+
+impl Into<RpcError> for DasApiError {
```
danenbm, line 19: If you implement the `From` trait instead of the `Into` trait, I believe errors can be converted implicitly **without** having to use  `map_err(Into::into)` as I see you are using in `api.rs`.  Basically `from()` will automatically be used to do the conversion.

```
impl From<DasApiError> for RpcError {
    fn from(error: DasApiError) -> Self
       ...
    }
}

File: digital_asset_types/src/rpc/asset.rs (hunk 4 above)
Comment (austbot) [reply digital_asset_types/src/rpc/asset.rs:34]: yes

File: nft_ingester/src/parsers/bubblegum.rs (hunk 6 above)
Comment (austbot) [reply nft_ingester/src/parsers/bubblegum.rs:80]: nope writes to the db other than ephemeral

File: nft_ingester/src/parsers/bubblegum.rs (hunk 5 above)
Comment (austbot) [reply nft_ingester/src/parsers/bubblegum.rs:81]: death by 10000 websites

File: digital_asset_types/src/dao/asset.rs [hunk 8]
```diff
@@ -97,8 +101,8 @@ impl ColumnTrait for Column {
             Self::RoyaltyTarget => ColumnType::Binary.def().null(),
             Self::RoyaltyAmount => ColumnType::Integer.def(),
             Self::ChainDataId => ColumnType::BigInteger.def().null(),
-            Self::CreatedAt => ColumnType::Timestamp.def().null(),
-            Self::BurntAt => ColumnType::Timestamp.def().null(),
+            Self::CreatedAt => ColumnType::TimestampWithTimeZone.def().null(),
```
danenbm, line 104: Do you want this to be with timezone or without?  When I generated it it defaulted to with timezone, not sure if you had chosen that vs utc and why but wanted to point it out in case it reverted by mistake.

nft_ingester/src/tasks/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl TaskManager {
+    pub fn new(name: String, pool: Pool<Postgres>) -> Result<Self, IngesterError> {
+        let runtime = Builder::new_multi_thread()
+            .enable_all()
+            .thread_name(name)
+            .build()
+            .map_err(|err| IngesterError::TaskManagerError(format!("Could not create tokio runtime: {:?}", err)))?;
+
+        let (producer, mut receiver) = mpsc::unbounded_channel::<Box<dyn BgTask>>();
```
danenbm, line 29: Cool use of tokio channel!

PR #118, opened 2022-06-22
Project canopy

jarry-xiao:
Project canopy increases the potential Merkle tree size by an order of magnitude by allowing Gummyroll to store more hashes on chain. The main problem that this solves is that it enables users to supply proofs while passing in fewer hashes into the transaction (which is limited to 1232 bytes).

This change can support the ability to index and operate on trees with significantly more leaves than was previously possible with the added benefit of not modifying any existing APIs

### DIFF
--- a/contracts/programs/bubblegum/src/lib.rs
+++ b/contracts/programs/bubblegum/src/lib.rs
@@ -13,7 +13,6 @@ use {
     },
     gummyroll::{program::Gummyroll, Node},
     crate::utils::{append_leaf,
-                   insert_or_append_leaf,
                    replace_leaf,
                    get_asset_id,
     },
@@ -567,7 +566,7 @@ pub mod bubblegum {
         }?;
         let merkle_slab = ctx.accounts.merkle_slab.to_account_info();
         emit!(voucher.leaf_schema.to_event());
-        insert_or_append_leaf(
+        replace_leaf(
             &merkle_slab.key(),
             *ctx.bumps.get("authority").unwrap(),
             &ctx.accounts.gummyroll_program.to_account_info(),
@@ -575,6 +574,7 @@ pub mod bubblegum {
             &ctx.accounts.merkle_slab.to_account_info(),
             ctx.remaining_accounts,
             root,
+            [0; 32], 
             voucher.leaf_schema.to_node(),
             voucher.index,
         )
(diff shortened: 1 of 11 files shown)


contracts/programs/bubblegum/src/lib.rs
```diff
@@ -567,14 +566,15 @@ pub mod bubblegum {
         }?;
         let merkle_slab = ctx.accounts.merkle_slab.to_account_info();
         emit!(voucher.leaf_schema.to_event());
-        insert_or_append_leaf(
+        replace_leaf(
```
jarry-xiao, line 569: I think it makes sense for `cancel_redeem` to only attempt to replace given that the nonce (index) is known

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -358,33 +322,30 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+    ) -> Result<bool, CMTError> {
         let mask: usize = MAX_BUFFER_SIZE - 1;
-        let changelog_buffer_index: u64;
-        if use_full_buffer {
-            changelog_buffer_index = self.active_index.wrapping_sub(self.buffer_size) & mask as u64
-        } else {
-            match self.find_root_in_changelog(current_root) {
-                Some(matching_changelog_index) => {
-                    changelog_buffer_index = matching_changelog_index;
+        let (changelog_buffer_index, use_full_buffer) = match self
```
jarry-xiao, line 335: `use_full_buffer` should be determined, not specified

contracts/programs/gummyroll/src/lib.rs
```diff
@@ -58,70 +60,196 @@ pub struct TransferAuthority<'info> {
...
+        // node_idx - 2 maps to the canopy index
+        let shifted_index = node_idx as usize - 2;
+        if shifted_index % 2 == 0 {
+            inferred_nodes.push(canopy[shifted_index + 1])
+        } else {
+            inferred_nodes.push(canopy[shifted_index - 1])
+        }
+        node_idx >>= 1;
+    }
+    proof.extend(inferred_nodes.iter());
```
jarry-xiao, line 160: The correct behavior here is probably to only extend if the proof node is not already provided. Will add the change in soon

samwise2 (approved):
lgtm! super helpful for marketplace 🎉

PR #121, opened 2022-06-24
Jxiao/fix log truncation

jarry-xiao:
Absolutely brutal bug and disgusting hack


contracts/programs/candy-wrapper/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    Ok(())
+}
+
+pub fn wrap_instruction(data: Vec<u8>) -> Instruction {
+    Instruction {
+        program_id: crate::id(),
+        accounts: vec![],
+        data,
+    }
+}
```
ngundotra, line 25: haha smart contract engineering too ez
💀 💀

contracts/programs/gummyroll/src/lib.rs
```diff
@@ -296,6 +299,7 @@ pub mod gummyroll {
         let (roll_bytes, canopy_bytes) = rest.split_at_mut(merkle_roll_size);
         let id = ctx.accounts.merkle_roll.key();
         let change_log = merkle_roll_apply_fn!(header, id, roll_bytes, initialize,)?;
+        wrap_event(change_log.try_to_vec()?, &ctx.accounts.candy_wrapper)?;
         emit!(*change_log);
```
ngundotra, line 303: Just restating here:

The point of having BOTH event data in logs & in CPI args saved to transaction information is to allow for log-based indexing, correct?

i.e. if logs are truncated for appends, we would have to replay the transaction(s) to figure out the order in which the new NFTs were appended. However this requires maintaining state while parsing (to figure out the order of appends), which is undesirable at scale. This is why we emit changelogs in the WRAP cpi args -- to allow for stateless parsing, correct?

contracts/sdk/gumball-machine/instructions/index.ts
```diff
@@ -39,117 +35,142 @@ export async function createInitializeGumballMachineIxs(
...
+    allocGumballMachineAcctInstr,
+    allocMerkleRollAcctInstr,
+    initGumballMachineInstr,
+  ];
 }
 
 /**
  * Wrapper on top of Solita's createDispenseNftSolInstruction. Automatically fetches necessary PDA keys for instruction
  * */
 export async function createDispenseNFTForSolIx(
```
ngundotra, line 93: @samwise2 great job setting these up! They came in handy :)

ngundotra (approved):
Pretty clean change over all.

Left a comment about indexing so the purpose of the CPI into WRAP program is more obvious. cc @austbot @danenbm for comment

contracts/programs/gummyroll/src/lib.rs
jarry-xiao: The rationale is that we don't want the existing system to break just yet. Common keep both paradigms and deprecate when everything has been migrated

austbot: +1

PR #122, opened 2022-06-27
[WIP] Better Backfiller

ngundotra:
Backfilling logic:

1. Repair existing db table

- Search for gaps in the db
- Query txIds for the tree address in between those gaps
- [New] Deduplicate the txId list by their slot
- [New] Query slots in batches of 20
- Parse query slots asynchronously

2. Check if on-chain tree is ahead of db
- Request txIds for address up to on-chain seq #
- [New] Deduplicate the txId list by their slot
- [New] Query slots in batches of 20
- Parse query slots asynchronously


contracts/sdk/indexer/db.ts
```diff
@@ -83,6 +100,35 @@ export class NFTDatabaseConnection {
...
+      DELETE FROM merkle
+      WHERE seq in 
+        (SELECT seq FROM 
+          (SELECT node_idx, MAX(seq) AS seq 
+          FROM merkle 
+          WHERE level = 0 group by node_idx
+        ));
+      `
+    );
+  }
```
ngundotra, line 130: This is used for testing the backfill scripts.

contracts/sdk/indexer/db.ts
```diff
@@ -657,6 +746,38 @@ export class NFTDatabaseConnection {
     return transactionId.length ? transactionId[0].transaction_id as string : null;
   }
 
+  async getAssetInfo(assetId: string): Promise<AssetInfo> {
```
ngundotra, line 749: This should be in a different PR. This is used by server

contracts/sdk/indexer/db.ts
jarry-xiao: Let's not commit this one lol. There's literally no use case except for testing

ngundotra: Okay fair point, I'd rather add to a readme

PR #129, opened 2022-06-30
Fix decompress mint seeds

ngundotra:
changed in previous PR causing `bubblegum-*` tests to break


ngundotra:
Actually no need to fix contract. Changing `mint` to be `assetId` was intended behavior.

PR #130, opened 2022-06-30
External Configuration

austbot:
In preparation for deploy to testnet I externalized the database and redis configuration to allow this to run in the cloud.

I also fixed an accounts ordering bug created with candy wrapper and a bug where the tests did not match the needed seeds for decompress.

ID == mint in NFT land on the v1 specification so I needed to make sure asset id is the passed in account for mint, and that mint authority (ephemeral) is a derived account from asset

### DIFF
--- a/contracts/programs/bubblegum/src/lib.rs
+++ b/contracts/programs/bubblegum/src/lib.rs
@@ -69,14 +69,12 @@ pub struct CreateTree<'info> {
 
 #[derive(Accounts)]
 pub struct MintV1<'info> {
-    /// CHECK: This account is neither written to nor read from.
     pub mint_authority: Signer<'info>,
     #[account(
+    mut,
     seeds = [merkle_slab.key().as_ref()],
     bump,
     )]
-    /// CHECK: This account is neither written to nor read from.
-    #[account(mut)]
     pub authority: Account<'info, Nonce>,
     pub candy_wrapper: Program<'info, CandyWrapper>,
     pub gummyroll_program: Program<'info, Gummyroll>,
@@ -95,7 +93,6 @@ pub struct Burn<'info> {
     seeds = [merkle_slab.key().as_ref()],
     bump,
     )]
-    /// CHECK: This account is neither written to nor read from.
     pub authority: Account<'info, Nonce>,
     pub candy_wrapper: Program<'info, CandyWrapper>,
     pub gummyroll_program: Program<'info, Gummyroll>,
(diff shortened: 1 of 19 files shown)


docker/runs.sh
```diff
@@ -53,11 +53,11 @@ args=(
   --rpc-port 8899
   --bpf-program metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s /so/mpl_token_metadata.so
   --bpf-program BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY /so/bubblegum.so
-  --bpf-program Fg6PaFpoGXkYsidMpWTK6W2BeZ7FEfcYkg476zPFsLnS /so/gummyroll_crud.so
```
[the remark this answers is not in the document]
danenbm: Also I think we need to remove gummyroll_crud from the Anchor.toml.  If its working for you now its probably because the old .so and idl files are still there from previous.

plerkle/src/geyser_plugin_nft.rs
```diff
@@ -188,10 +201,18 @@ impl<T: 'static + Messenger + Default + Send + Sync> GeyserPlugin for Plerkle<'s
...
+            .map_err(|config_error|
+                GeyserPluginError::ConfigFileReadError {
+                    msg: format!("Could not read messenger config: {:?}", config_error)
+                }
+            )?;
         runtime.spawn(async move {
             // Create new Messenger connection.
-            if let Ok(mut messenger) = T::new().await {
+
+            if let Ok(mut messenger) = T::new(config.messenger_config).await {
```
danenbm, line 215: This is really cool, messenger stays generic because the config is generic, and its really simple to add the environment stuff in the docker yaml.

ngundotra:
Thanks for making test change!

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -95,16 +93,15 @@ pub struct Burn<'info> {
...
     pub authority: Account<'info, Nonce>,
     pub candy_wrapper: Program<'info, CandyWrapper>,
     pub gummyroll_program: Program<'info, Gummyroll>,
     /// CHECK: This account is checked in the instruction
     pub owner: UncheckedAccount<'info>,
     /// CHECK: This account is chekced in the instruction
     pub delegate: UncheckedAccount<'info>,
+
     #[account(mut)]
-    /// CHECK: unsafe
```
danenbm, line 107: Doesn't this need to be here for `merkle_slab`?

austbot: yeah weird not sure why those are gone

docker/runs.sh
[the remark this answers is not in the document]
austbot: good call will remove

PR #131, opened 2022-07-01
Adding backfiller task to ingester

danenbm:
### Adding backfiller task to ingester
- Design based on existing `indexer.ts`, `backfiller.ts`, and `backfill.ts`
but using postgres database and geyser-based forward filling.
- Adding `backfill_items` table to postgresql database.
- Also adding trigger to send notification when items are inserted
into `backfill_items` table.
- Task gets current trees listed in `backfill_items` that have multiple
rows or `force_chk` flag set.
- Task calls unimplemented functions to fill gaps for each case.
- Task deletes extra rows from the `backfill_items` table. (One row
for each tree must remain to detect future gaps).

### Testing
- Ran bubblegum-test and observed `backfill-items` get appended to,
trigger the backfill task, and then the extra rows removed.
- Instrumented code to add cases with `force_chk` flag set and
observed backfiller run the correct code to backfill in that case.
- Instrumented code to add erroneous duplicates to the table and
observed the code behave correctly and remove the duplicates when
completing the backfill.


init.sql
```diff
@@ -25,6 +25,38 @@ CREATE INDEX cl_items_node_idx on cl_items (node_idx);
 CREATE INDEX cl_items_leaf_idx on cl_items (leaf_idx);
 CREATE UNIQUE INDEX cl_items__tree_node on cl_items (tree, node_idx);
 
+CREATE TABLE backfill_items
+(
+    id          bigserial PRIMARY KEY,
+    tree        BYTEA  NOT NULL,
+    seq         BIGINT NOT NULL,
+    slot        BIGINT NOT NULL,
+    force_chk   bool
```
ngundotra, line 34: Why store `id` in this table? Seems like this is just used to track gaps in the `tree` sequence numbers?

ngundotra (commented):
Left some comments. Looks good.

init.sql
danenbm: I agree I do not think ID is needed for anything, I was just following the fact that all of the other tables had an ID as a primary key.

nft_ingester/src/main.rs
```diff
@@ -57,9 +65,19 @@ async fn main() {
...
+    tasks.push(
+        service_transaction_stream::<RedisMessenger>(
+            pool.clone(),
+            background_task_manager.get_sender(),
+            config.messenger_config.clone(),
+        )
+        .await,
+    );
+    // Start up backfiller process.
+    tasks.push(backfiller(pool.clone()).await);
```
austbot, line 80: nice

nft_ingester/src/main.rs
```diff
@@ -116,6 +134,202 @@ async fn service_account_stream<T: Messenger>(
...
+                    // Print error but keep trying.
+                    println!("Could not get trees from db: {err}");
+                }
+            }
+        }
+    })
+}
+
+#[derive(Debug, FromQueryResult)]
+struct UniqueTree {
```
austbot, line 220: we can do this in a new PR but we prolly dont need all the backfiller logic in main, we can move it to a package

nft_ingester/src/main.rs
```diff
@@ -116,6 +134,202 @@ async fn service_account_stream<T: Messenger>(
...
+        r#"SELECT DISTINCT backfill_items.tree FROM backfill_items WHERE backfill_items.force_chk = TRUE"#,
+        vec![],
+    ))
+    .all(db)
+    .await?;
+
+    // Get trees with multiple rows from `backfill_items` table.
+    let multi_row_trees = UniqueTree::find_by_statement(Statement::from_sql_and_values(
+        DbBackend::Postgres,
+        r#"SELECT backfill_items.tree FROM backfill_items GROUP BY backfill_items.tree HAVING COUNT(*) > 1"#,
```
austbot, line 242: SELECT count(*) as c ... group by .. HAVING c > 1 ,

Is this the same thing? curious if one sucks speed wise and the other is super fast

nft_ingester/src/main.rs
```diff
@@ -116,6 +134,202 @@ async fn service_account_stream<T: Messenger>(
...
+    .await?;
+
+    Ok(BackfillTrees {
+        force_chk_trees,
+        multi_row_trees,
+    })
+}
+
+async fn backfill_tree_from_seq_1(_db: &DatabaseConnection, _tree: &[u8]) -> Result<(), DbErr> {
+    //TODO implement gap filler.
```
austbot, line 255: best code in the project

nft_ingester/src/parsers/gummyroll.rs
```diff
@@ -154,6 +163,44 @@ pub async fn gummyroll_change_log_event_to_database(
             .await
             .map_err(|db_err| IngesterError::StorageWriteError(db_err.to_string()))?;
     }
+
+    // If and only if the entire path of nodes was inserted into the `cl_items` table, then insert
```
austbot, line 167: nice catch this will be self healing for tansient db faiilures on insert

austbot (approved):
Love it and I like it

contracts/programs/bubblegum/src/error.rs
```diff
@@ -10,4 +10,4 @@ pub enum BubblegumError {
     HashingMismatch,
     #[msg("Unsupported Schema Version")]
     UnsupportedSchemaVersion,
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
danenbm: I think it was just a whitespace change.

init.sql
```diff
@@ -25,6 +25,38 @@ CREATE INDEX cl_items_node_idx on cl_items (node_idx);
...
+);
+
+CREATE INDEX backfill_items_tree_idx on backfill_items (tree);
```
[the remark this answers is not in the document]
danenbm: wdyt about:
```
CREATE INDEX backfill_items_tree_seq_idx on cl_items (tree, seq);
CREATE INDEX backfill_items_tree_slot_idx on cl_items (tree, slot);
```

nft_ingester/src/main.rs
danenbm: Thanks!

nft_ingester/src/main.rs
danenbm: lol

nft_ingester/src/parsers/gummyroll.rs
danenbm: Thanks, I was trying to think about partial failures, duplicated records, and out of order records, since I'm not actually sure what kind of performance issues are possible yet in our overall system.

nft_ingester/src/main.rs
danenbm: What I have will give me a list of trees:
```
solana=# SELECT backfill_items.tree FROM backfill_items GROUP BY backfill_items.tree HAVING COUNT(*) > 1;
                                tree                                
--------------------------------------------------------------------
 \x3aa91498110d743f0b0b510564b26862b252dfd3290059bbcf49fc301184990a
(1 row)
```

I tried to do the query you suggest and it doesn't like `c` in the HAVING expression.:
```
solana=# SELECT COUNT(*) as c GROUP BY backfill_items.tree HAVING c > 1;
ERROR:  column "c" does not exist
LINE 1: ...ECT COUNT(*) as c GROUP BY backfill_items.tree HAVING c > 1;
                                                                 ^
```

But `SELECT COUNT(*)` will return a count and at this spot in the code I'm looking for a tree that has the count:
```      
solana=# SELECT COUNT(*) FROM backfill_items GROUP BY backfill_items.tree HAVING COUNT(*) > 1;
 count 
-------
     7
```

Was that the query you meant or did I misunderstand?  Overall I'm trying to get distinct tree IDs for which there are multiple rows in the database.

init.sql
```diff
@@ -25,6 +25,38 @@ CREATE INDEX cl_items_node_idx on cl_items (node_idx);
...
+CREATE INDEX backfill_items_seq_idx on backfill_items (seq);
+
+CREATE or REPLACE FUNCTION notify_new_backfill_item()
+    RETURNS trigger
+     LANGUAGE 'plpgsql'
+as $BODY$
+declare
+begin
+    if (tg_op = 'INSERT') then
```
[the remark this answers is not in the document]
danenbm: So far from my basic research it seems like triggering procedures cause issues if you are incorrectly doing complex relational integrity things and modifying your data.  Triggers are global and they always run regardless of circumstances.  Also nested procedures, recursive ones, and similar, get complicated and hard to debug.  However, it still seems like some high performance database management systems use them for things like auto-generated columns.

But my use case is exceedingly simple.  I'm not modifying any data with this, or even counting on this event to refer to a specific INSERT event.  There's no data integrity implications as I'm using the trigger as a semaphore to wake up a sleeping task.

It seems like it would just be a question of overall load performance issue for the database, and since this procedure is so minimal (it just calls a single `pg_notify()` method), I'm doubting it will bog the system down in any way.

Plus, even if we miss an event, we will just be triggered next time as well.  There's no coupling between the trigger and the trees to backfill.  I will keep looking into this topic more.

nft_ingester/src/main.rs
danenbm: Created issue in Linear, PL-150:

Move backfiller logic to its own crate
Move backfiller out of main ingester code.  Also move database queries into correct location in digital_asset_types crate.

init.sql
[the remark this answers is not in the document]
austbot: Sounds reasonable it's something i havent done with postgress but with rethink and mongo db /lgtm

nft_ingester/src/main.rs
austbot: I was just curious i having clause can pull from preselected /lgtm

austbot (approved):
/lgtm very reasonable

PR #132, opened 2022-07-04
Fix Gumball Machine Dispense Logic

samwise2:
We were previously always charging users for only one NFT. This is now fixed. We also log a `msg!` when a request to mint occurs when there are no more remaining NFTs for mint, since this seemed like something that would be useful for off-chain indexing.


samwise2:
merging w offline approval from @jarry-xiao

PR #133, opened 2022-07-05
Gumball Machine: Add Creators to Header

samwise2:
We want to add creators to the `GumballMachineHeader` to enable projects to set a list of secondary sale royalty recipients (rather than just one). 

This PR enables projects to set up to `NUM_CREATORS (=5)` creators in their header which will then be set as an immutable list of creators on compressed NFTs that are dispensed from their machine.


contracts/programs/gumball-machine/src/state/mod.rs
```diff
@@ -28,6 +28,43 @@ impl EncodeMethod {
...
+    pub fn adapt(&self) -> bubblegum::state::metaplex_adapter::Creator {
+        bubblegum::state::metaplex_adapter::Creator {
+            address: self.address,
+            verified: self.verified == 1,
+            share: self.share,
+        }
+    }
+    pub fn is_valid(&self) -> bool {
+        return self.address != Default::default() && self.share != 0
+    }
```
ngundotra, line 65: `self.share > 0`

contracts/programs/gumball-machine/src/lib.rs
```diff
@@ -318,12 +318,29 @@ pub mod gumball_machine {
...
         let (mut header_bytes, config_data) =
             gumball_machine_data.split_at_mut(std::mem::size_of::<GumballMachineHeader>());
         let gumball_header = GumballMachineHeader::load_mut_bytes(&mut header_bytes)?;
         let size = max_items as usize;
+
+        // Construct creators array
+        let mut creators: [Creator; NUM_CREATORS] = [Default::default(), Default::default(), Default::default(), Default::default(), Default::default()];
+        assert_eq!(creator_keys.len(), creator_shares.len());
+        assert!(creator_keys.len() < NUM_CREATORS, "Cannot set more than {} creators", NUM_CREATORS);
+        assert!(creator_shares.iter().sum::<u8>() <= 100, "Cannot have creator royalty percentages total more than 100% of the sale price");
```
ngundotra, line 334: I think adding a check for ` ... == 100` is useful, because we also don't want over 100 percent either

ngundotra (commented):
Where are you testing that the creators are receiving secondary royalty sales?

samwise2:
> Where are you testing that the creators are receiving secondary royalty sales?

gumball machine doesn't provide any secondary sale functionality so this isn't tested in the gumball machine units, but we do test that the creators are set as expected in the header (and via assertions I checked that they're set correctly in the CPI to `Bubblegum::mint`). 

Sugar shack tests that we pay royalties correctly according to the creators vec [link] that the NFT was minted with in Bubblegum, so that would be where this is tested. At some point a Gumball Machine + Sugar Shack integration suite might make sense.

ngundotra:
That's right! Makes sense. Lgtm

ngundotra (approved):

contracts/programs/gumball-machine/src/lib.rs
```diff
@@ -336,12 +336,29 @@ pub mod gumball_machine {
...
+        creator_shares: Vec<u8>
     ) -> Result<()> {
         let mut gumball_machine_data = ctx.accounts.gumball_machine.try_borrow_mut_data()?;
         let (mut header_bytes, config_data) =
             gumball_machine_data.split_at_mut(std::mem::size_of::<GumballMachineHeader>());
         let gumball_header = GumballMachineHeader::load_mut_bytes(&mut header_bytes)?;
         let size = max_items as usize;
+
+        // Construct creators array
+        let mut creators: [Creator; NUM_CREATORS] = [Default::default(), Default::default(), Default::default(), Default::default(), Default::default()];
```
jarry-xiao, line 349: can you run `cargo fmt` on the whole gumball-machine module?

contracts/programs/gumball-machine/src/state/mod.rs
```diff
@@ -28,6 +28,43 @@ impl EncodeMethod {
     }
 }
 
+pub const NUM_CREATORS: usize = 5;
+
+// Unfortunately, duplicate so that Creator is picked up by Solita
+#[repr(C)]
+#[derive(AnchorSerialize, AnchorDeserialize, PartialEq, Copy, Clone, Zeroable, Pod)]
+pub struct Creator {
```
jarry-xiao, line 36: Maybe rename this so we know it's an adapter class

PR #137, opened 2022-07-07
Cleaned up CMT lib by refactoring away a couple common functions

jarry-xiao:
- Created a function called `traverse_to_parent` to abstract away the ugly (and highly duplicated) code that hashes a reference node up the tree
- Created a function inside the `ChangeLog` class that handles proof node or leaf replacement. If this existed months ago, we would have never introduced the bug in replacing the RML
- Use `empty_node_cached` when possible (O(N^2) -> O(N)) preprocessing
- Refactored `append` to be more easily readable
- Moved most direct assignment of variables into functions

All test cases pass and smoke test is chugging along

### DIFF
--- a/lib/concurrent-merkle-tree/src/merkle_roll.rs
+++ b/lib/concurrent-merkle-tree/src/merkle_roll.rs
@@ -1,17 +1,16 @@
 use crate::{
     error::CMTError,
     state::{ChangeLog, Node, Path, EMPTY},
-    utils::{empty_node, fill_in_proof, recompute},
+    utils::{empty_node, empty_node_cached, fill_in_proof, recompute, traverse_to_parent},
 };
 use bytemuck::{Pod, Zeroable};
 pub(crate) use log_compute;
 pub(crate) use solana_logging;
 
-use solana_program::keccak::hashv;
-
 #[cfg(feature = "sol-log")]
 use solana_program::{log::sol_log_compute_units, msg};
 
+#[inline(always)]
 fn check_bounds(max_depth: usize, max_buffer_size: usize) {
     assert!(max_depth < 31);
     // This will return true if MAX_BUFFER_SIZE is a power of 2 or if it is 0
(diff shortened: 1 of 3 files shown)


lib/concurrent-merkle-tree/src/utils.rs
```diff
@@ -31,19 +25,23 @@ pub fn empty_node_cached<const N: usize>(level: u32, cache: &mut Box<[Node; N]>)
...
+    for (depth, sibling) in proof.iter().enumerate() {
+        traverse_to_parent(&mut current_node, sibling, index >> depth & 1 == 0);
     }
-
     current_node
 }
 
+/// Computes the parent node of `node` and `sibling` and copies the result into `node`
+#[inline(always)]
+pub fn traverse_to_parent(node: &mut Node, sibling: &Node, is_left: bool) {
```
ngundotra, line 36: i think `hash_to_parent` is a better name

lib/concurrent-merkle-tree/src/state.rs
```diff
@@ -23,27 +23,46 @@ impl<const MAX_DEPTH: usize> ChangeLog<MAX_DEPTH> {
...
-                start.copy_from_slice(res.as_ref());
-            } else {
-                let res = hashv(&[s.as_ref(), start.as_ref()]);
-                start.copy_from_slice(res.as_ref());
-            }
-            if ix < MAX_DEPTH - 1 {
-                self.path[ix + 1] = start;
-            }
+    pub fn recompute_path(&mut self, index: u32, mut node: Node, proof: &[Node]) -> Node {
+        self.index = index;
```
ngundotra, line 41: why does recomputing path set the `index` of the leaf?

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -395,43 +366,34 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
-                    let critbit_index = (MAX_DEPTH - 1) - common_path_len;
-                    self.rightmost_proof.proof[critbit_index] = change_log.path[critbit_index];
-                } else {
-                    self.rightmost_proof.leaf = change_log.get_leaf();
-                }
+                change_log.update_proof_or_leaf(
+                    self.rightmost_proof.index - 1,
+                    &mut self.rightmost_proof.proof,
+                    &mut self.rightmost_proof.leaf,
+                );
```
ngundotra, line 396: this is the most useful part of the PR 👍

lib/concurrent-merkle-tree/src/state.rs
```diff
@@ -23,27 +23,46 @@ impl<const MAX_DEPTH: usize> ChangeLog<MAX_DEPTH> {
...
+        proof: &mut [Node; MAX_DEPTH],
+        leaf: &mut Node,
+    ) {
+        let padding: usize = 32 - MAX_DEPTH;
+        if leaf_index != self.index {
+            // This bit math is used to identify which node in the proof
+            // we need to swap for a corresponding node in a saved change log
+            let common_path_len = ((leaf_index ^ self.index) << padding).leading_zeros() as usize;
+            let critbit_index = (MAX_DEPTH - 1) - common_path_len;
+            proof[critbit_index] = self.path[critbit_index];
```
ngundotra, line 62: it would be nice to refactor this using normal modular addition 🙃

ngundotra (commented):
This PR would be better if we could refactor the remaining bit math to normal modular arithmetic.

Also, what does `Moved most direct assignment of variables into functions` mean?

lib/concurrent-merkle-tree/src/state.rs
jarry-xiao: you're updating the change log

lib/concurrent-merkle-tree/src/state.rs
jarry-xiao: I don't think there's any loss in safety or even any loss to the end user to do things this way. Every mutable public function calls `check_bounds`

ngundotra (approved):

PR #139, opened 2022-07-08
Gumball Machine: Initialize Indices in Chunks

samwise2:
A computational bottleneck in the GumballMachine program is initializing the fisher yates indices which are used to provide on-chain randomness for minting NFTs. This initialization requires is O(`max_items`) in time, which is problematic for projects at large scale (we run out of compute trying to initialize all indices on the gumball machine account in one go). For this reason, we've chunked the initialization into segments, so that all of the indices can be initialized via multiple transactions. 

Indices are initialized in chunks of at most `250,000` unites, to stay within compute budget. 

We've added a CLI instruction that will automatically send n transactions to the network to fully initialize all indices for a gumball machine (so the user doesn't have to concern themselves with the chunking mechanism).

We've also added support for gumball machines without config lines. By setting `extension_len` to 0, and fully initializing your indices, you enable users to mint NFTs from your machine without config lines. You could then use an off-chain mapping of NFT index -> arweave identifier, etc. to provide meaning to these config lines.


contracts/programs/gumball-machine/src/state/mod.rs
```diff
@@ -99,8 +99,11 @@ pub struct GumballMachineHeader {
     pub extension_len: u64,
     pub max_mint_size: u64,
     pub remaining: u64,
+    // why is max_items a u64 when our indices are written as u32s...this could break, our index writing could overflow
+    // we should bound it to 32 or write indices as u64s
     pub max_items: u64,
```
samwise2, line 104: Wanted to raise this. The indices we write to the gumball machine account are size u32. But we allow max_items to have size bounded by u64. Since we write 0...`max_items` as indices on account, this could cause overflow. For now I set `smallest_uninitialized_index` to be a u64 to match. But we should probably be careful about this

jarry-xiao: This is a good point. We can make this a u32, but need to byte align the struct

contracts/programs/gumball-machine/src/lib.rs
```diff
@@ -264,8 +278,17 @@ fn find_and_mint_compressed_nfts<'info>(
     let (mut header_bytes, config_data) =
         gumball_machine_data.split_at_mut(std::mem::size_of::<GumballMachineHeader>());
     let gumball_header = GumballMachineHeader::load_mut_bytes(&mut header_bytes)?;
+
+    // Cannot dispense before all indices are initialized
+    assert_eq!(gumball_header.max_items, gumball_header.smallest_uninitialized_index);
+
+    // Cannot dispense more than the max_mint_size
+    assert!(num_items <= gumball_header.max_mint_size);
```
samwise2, line 286: @jarry-xiao I was sort of guessing what this was supposed to be used for lol. Is this what you had in mind?

jarry-xiao: yeah I think this looks reasonable

contracts/programs/gumball-machine/src/lib.rs
```diff
@@ -199,25 +207,31 @@ fn fisher_yates_shuffle_and_fetch_nft_metadata<'info>(
...
-        config_lines_data[random_config_index..random_config_index + line_size].to_vec();
-
-    let nft_index = (random_config_index / line_size) + 1;
+    let zero_nft_index = indices[(gumball_header.remaining as usize)] as usize;
+    let nft_index = zero_nft_index + 1;
+    msg!("Minted NFT with 1-index {}", nft_index);
+    let random_config_index = zero_nft_index * line_size;
+    let config_line = if config_lines_data.len() > 0 {
+                          // If the machine is manually specifying config lines then extract the config line data
+                          config_lines_data[random_config_index..random_config_index + line_size].to_vec()
```
jarry-xiao, line 230: please run `cargo fmt` on this module

contracts/programs/gumball-machine/src/lib.rs
```diff
@@ -426,6 +446,42 @@ pub mod gumball_machine {
...
+
+        // Grab mutable reference to indices bytes
+        let size = gumball_header.max_items as usize;
+        let index_array_size = std::mem::size_of::<u32>() * size;
+        let (indices_data, _) = config_data.split_at_mut(index_array_size);
+        let indices = cast_slice_mut::<u8, u32>(indices_data);
+
+        // Determine the next byte range to initialize
+        let first_index_to_initialize = gumball_header.smallest_uninitialized_index as usize;
+        let next_smallest_uninitialized_index = (gumball_header.smallest_uninitialized_index + 250000).min(gumball_header.max_items) as usize;
```
jarry-xiao, line 470: not a huge fan of the hardcoded 250000

jarry-xiao (approved):
Minor formatting suggestions, but feel free to merge once those are in

contracts/programs/gumball-machine/src/lib.rs
samwise2: i'll pull this into `state` as a constant

PR #142, opened 2022-07-12
Add limit and offset to jankle assets endpoint

ngundotra:
Paginate `/assets`  endpoint


ngundotra:
Server returns 3000+ assets... feels bad. Easier to let front end  decide what to ask for

jarry-xiao:
> Server returns 3000+ assets... feels bad. Easier to let front end decide what to ask for

How does a single user have 3000+ assets?????

PR #146, opened 2022-07-13
Added more metadata columns to db

jarry-xiao:


ngundotra:
Can you also add `compressed DESC, name` to the order by clause?

ngundotra (changes requested):
see comments

PR #150, opened 2022-07-19
Adding seq number to Plerkle asset tables (Old PR)

danenbm:
### This PR was replaced with https://github.com/jarry-xiao/candyland/pull/158


danenbm:
One suggestion was to solve this by emit/wrapping the new leaf schema for any leaf-modifying operation.  To me, that  seems the easiest method that fits into what we currently are doing.

Other suggestions (I still need to understand more) involve always appending to the `cl_items` table and potentially doing gap management natively in the database.

PR #153, opened 2022-07-19
[Gummyroll] Single authority for each tree

ngundotra:
Use a single authority for each tree in order to simplify composability.


ngundotra:
All `anchor test` cases pass

PR #154, opened 2022-07-20
Support Decompress with Creators

samwise2:
We had an issue where we could not decompress NFTs with multiple creators specified. `MPL` expects one of the creators (if present) to be the `update_authority` for the metadata account associated with the decompressed NFT. This `update_authority` must also sign in the creation of the master edition. Since `create_metadata_accounts_v2`, `create_master_edition_v3` are called from `Bubblegum` via CPI, we needed to force a signer for `update_authority`. Thus, if the user has specified a non-zero number of creators, then we forcibly append the `mint_authority` PDA owned by `Bubblegum` as the last creator -- which is also the `update_authority` in the CPI call, thus enabling decompress.

Limitations of this approach:
- Since the `update_authority` is a PDA of `Bubblegum`, `Bubblegum` must provider wrappers around any instructions which must be signed by `update_authority` (other than `create_master_edition_v3`, since that occurs within `decompress`). For example, `Bubblegum` may need to provide a wrapper around `update_metadata_accounts`.

Originally, I thought that was a bigger deal, but it actually might not be such a bad thing. Updating metadata accounts lets some sort of project admin change the values in these structs held in the metadata account (an update would need to be run for each token I guess):

```rust
pub struct Data {
    /// The name of the asset
    pub name: String,
    /// The symbol for the asset
    pub symbol: String,
    /// URI pointing to JSON representing the asset
    pub uri: String,
    /// Royalty basis points that goes to creators in secondary sales (0-10000)
    pub seller_fee_basis_points: u16,
    /// Array of creators, optional
    pub creators: Option<Vec<Creator>>,
}

pub struct Metadata {
    pub key: Key,
    pub update_authority: Pubkey,
    pub mint: Pubkey,
    pub data: Data,
    // Immutable, once flipped, all sales of this metadata are considered secondary.
    pub primary_sale_happened: bool,
    // Whether or not the data struct is mutable, default is not
    pub is_mutable: bool,
    /// nonce for easy calculation of editions, if present
    pub edition_nonce: Option<u8>,
    /// Since we cannot easily change Metadata, we add the new DataV2 fields here at the end.
    pub token_standard: Option<TokenStandard>,
    /// Collection
    pub collection: Option<Collection>,
    /// Uses
    pub uses: Option<Uses>,
}
```

If a project admin knows that they want to update this data for all NFTs in the project, they would have to run this type of update for both compressed and non-compressed NFTs and hence having all of those updates going through Bubblegum doesn't seem too bad (and is also probably a super rare and expensive use case of this protocol in general).


contracts/tests/bubblegum-test.ts
```diff
@@ -406,14 +377,132 @@ describe("bubblegum", () => {
...
+      await execute(Bubblegum.provider, [mintIx], [payer]);
+
+      // Compute data hash
+      const metadataArgsBuffer = mintIx.data.slice(8)
+      const metadataArgsHash = keccak_256.digest(metadataArgsBuffer);
+      const sellerFeeBasisPointsNumberArray = bufferToArray(num16ToBuffer(metadata.sellerFeeBasisPoints))
+      const allDataToHash = metadataArgsHash.concat(sellerFeeBasisPointsNumberArray)
+      const dataHash = bufferToArray(
+        Buffer.from(keccak_256.digest(allDataToHash))
+      );
```
ngundotra, line 425: Can you add this to the `bubblegum/convenience` sdk?

contracts/tests/bubblegum-test.ts
```diff
@@ -406,14 +377,132 @@ describe("bubblegum", () => {
...
+
+      // Compute creator hash
+      let bufferOfCreatorData = Buffer.from([]);
+      let bufferOfCreatorShares = Buffer.from([]);
+      for (let creator of metadata.creators) {
+        bufferOfCreatorData = Buffer.concat([bufferOfCreatorData, creator.address.toBuffer(), Buffer.from([creator.share])])
+        bufferOfCreatorShares = Buffer.concat([bufferOfCreatorShares, Buffer.from([creator.share])])
+      }
+      let creatorHash = bufferToArray(Buffer.from(keccak_256.digest(bufferOfCreatorData)));
```
ngundotra, line 434: Can you add this to the `bubblegum/convenience` sdk?

contracts/tests/bubblegum-test.ts
```diff
@@ -406,14 +377,132 @@ describe("bubblegum", () => {
...
+          masterEdition: await getMasterEdition(asset),
+          sysvarRent: SYSVAR_RENT_PUBKEY,
+          tokenMetadataProgram: PROGRAM_ID,
+          associatedTokenProgram: ASSOCIATED_TOKEN_PROGRAM_ID,
+        },
+        {
+          metadata,
         }
       );
+      await execute(Bubblegum.provider, [decompressIx], [payer]);
```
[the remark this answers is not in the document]
ngundotra: The newest MPL js sdk might be able to help (but upgrading it to most recent version may break our project, so it's fine to just manually deserialize account & check)

ngundotra (changes requested):
Great work Sam. Thanks for cleaning up the test cases too.

Can you check the token metadata after decompressing & assert the creators are as expected?

ngundotra:
Also, can you rebase on top of #153  ?

contracts/sdk/bubblegum/src/convenience.ts
```diff
@@ -76,3 +77,40 @@ export async function getCreateTreeIxs(
     );
     return [allocAccountIx, initGummyrollIx];
 }
+
+export function computeMetadataArgsHash(mintIx: TransactionInstruction) {
+    const metadataArgsBuffer = mintIx.data.slice(8)
+    return keccak_256.digest(metadataArgsBuffer);
+}
+
+export function computeDataHash(sellerFeeBasisPoints: number, mintIx?: TransactionInstruction, metadataArgsHash?: number[]) {
```
samwise2, line 86: @ngundotra moved the `computeDataHash` and `computeCreatorHash` methods here. agree, way cleaner!

contracts/tests/bubblegum-test.ts
```diff
@@ -124,6 +124,70 @@ describe("bubblegum", () => {
...
+  };
+
+  const assertMetadataMatch = (onChainMetadata: Metadata, mintMetadataArgs: MetadataArgs, expectedMintAuthority: PublicKey) => {
+
+    const assertDataMatch = (onChainData: Data, expectedData: Data) => {
+      assert(trimStringPadding(onChainData.name) === expectedData.name, "names mismatched");
+      assert(trimStringPadding(onChainData.symbol) === expectedData.symbol, "symbols mismatched");
+      assert(trimStringPadding(onChainData.uri) === expectedData.uri, "uris mismatched");
+      assert(onChainData.sellerFeeBasisPoints === expectedData.sellerFeeBasisPoints)
+      onChainData.creators?.forEach((creator, index) => {
```
samwise2, line 161: asserting on matching creators (and the rest of the metadata)

contracts/programs/gumball-machine/src/state/mod.rs
```diff
@@ -97,13 +93,16 @@ pub struct GumballMachineHeader {
...
     pub max_items: u32,
     pub total_items_added: u32,
     pub smallest_uninitialized_index: u32,
+    // 0 for whitespace trimming, 1 for base58 encode
+    pub config_line_encode_method: u8,
+    // Secondary sale royalty recipients
+    pub creators: [GumballCreatorAdapter; NUM_CREATORS],
     // 8-byte align struct
-    pub _padding_2: [u8; 4],
+    pub _padding: [u8; 7],
```
samwise2, line 105: consolidated padding 🎉

samwise2:
Also added a fix to the gumball machine CLI to update gball header metadata with new params, and removed old `tests/utils` in favor of putting all utils in the sdk so they can be reused across CLI/indexer etc.

samwise2:
> Also, can you rebase on top of #153 ? We want to merge these two separately but both of these need to get in :)

Can totally rebase this, but it probably won't have the intended effect? i.e. if this gets merged in before #153 and is rebased on top of it, then some of the changes from #153 would come in with it (unless that's what you want?). Happy to wait for 153 to go first and then rebase on main once it's in? Or merge this and then 153 rebases on top. either works

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -748,7 +755,19 @@ pub mod bubblegum {
...
                 metadata.uri.clone(),
                 if metadata.creators.len() > 0 {
-                    Some(metadata.creators.iter().map(|c| c.adapt()).collect())
+                    let mut ammended_metadata_creators = metadata.creators;
+                    ammended_metadata_creators.push(Creator {
+                        address: ctx.accounts.mint_authority.key(),
+                        verified: true,
+                        share: 0,
+                    });
+                    assert!(ammended_metadata_creators.len() <= mpl_token_metadata::state::MAX_CREATOR_LIMIT, "Supplied metadata can have at most {} creators to leave space for mint_authority PDA creator", mpl_token_metadata::state::MAX_CREATOR_LIMIT-1);
```
ngundotra, line 764: Also throw this error on `mint_v1`.

contracts/programs/bubblegum/src/lib.rs
```diff
@@ -357,7 +361,10 @@ pub mod bubblegum {
 
         // @dev: seller_fee_basis points is encoded twice so that it can be passed to marketplace instructions, without passing the entire, un-hashed MetadataArgs struct
         let metadata_args_hash = keccak::hashv(&[message.try_to_vec()?.as_slice()]);
-        let data_hash = keccak::hashv(&[&metadata_args_hash.to_bytes(), &message.seller_fee_basis_points.to_le_bytes()]);
```
ngundotra, line 360: Right above this block, check that `creators` is valid.

Best way to do this is create `assert_data_valid` method like in `token_metadata`
[(link)](https://github.com/metaplex-foundation/metaplex-program-library/blob/cac9cfd042cc5a16f50d635074555392d3b9044e/token-metadata/program/src/utils.rs#L35-L132)

ngundotra (changes requested):
Great work! One last change & then I think it will be ready for merge after the gummyroll change

contracts/programs/bubblegum/src/lib.rs
samwise2: good catch, I was checking in reverse lol. We don't even need this assertion at all, since we know that MetadataArgs is compliant with whatever we allowed in `mint_v1` via the hash assertion at the top of this instruction. I'll build an `assert_data_valid` for `mint_v1` that ensures everything is MPL decompress compliant upfront, and then all we need to do is append the PDA in `decompress_v1`

contracts/programs/bubblegum/src/error.rs
```diff
@@ -10,4 +10,18 @@ pub enum BubblegumError {
     HashingMismatch,
     #[msg("Unsupported Schema Version")]
     UnsupportedSchemaVersion,
-}
\ No newline at end of file
+    #[msg("Creator shares must sum to 100")]
+    CreatorShareTotalMustBe100,
```
samwise2, line 14: needed to add custom errors because MPL token metadata errors don't work with Anchor

contracts/programs/bubblegum/src/utils.rs
```diff
@@ -8,6 +9,47 @@ use {
     gummyroll::Node,
 };
 
+/// Assert that the provided MetadataArgs are compatible with MPL `Data`
+pub fn assert_metadata_is_mpl_compatible(metadata: &MetadataArgs) -> Result<()> {
```
ngundotra, line 13: 👍

ngundotra (approved):
Great work Sam!

samwise2:
A few other things to note:
- this PR commits yarn.lock files as we look to transition over to yarn instead of npm
- this PR completely reshuffles a lot of the test files because of undefined behavior of async code in mocha `describe` blocks that was causing flake
- it eliminate test utils because duplicate versions of `execute` were causing problems and any helper methods the tests use could probably be used by a client side consumer

jarry-xiao (approved):
Contract changes look good to me

PR #160, opened 2022-07-24
Adding seq number to plerkle asset tables

danenbm:
### Notes
- This PR should work for what we have today, but now that the backfiller is another data source for the various database tables, we may want to revisit some of our strategies for `cl_items` and the various asset tables, since they can now get events out of order.

### Background
- This PR is needed to support replaying of Bubblegum instructions and avoid putting old data in various asset tables.
- Modifying `INSERT` statements for Bubblegum Mint instructions to have `ON_CONFLICT DO NOTHING` clauses.
 - This allows us to replay a `mint` instruction if there was a failure that caused some of the inserts to various tables to be missing.
- Modifying `UPDATE` statements for other Bubblegum instructions to have `WHERE` clauses that check `seq`.
  - There is/was a potential issue with this, because not every Bubblegum instruction writes the same fields to the asset tables.
  - **Problem:** If we always update the asset tables when we run old instructions during backfilling, then we might update newer data with old data.  On the other hand, If we always check the sequence number for the asset table before running the old instruction, we may miss updates from older instructions because the newer instructions might not have updated the exact same fields.
  - **Solution:** One suggestion was to always write all the relevant fields.  For example, have the Bubblegum `delegate` instruction processing also write `owner` to the database even though `owner` would not be changing during a `delegate` instruction.  The relevant fields are present in the leaf schema today.
  - Note the `burn` command and the `decompress` command cannot follow this idea for different reasons, but I don't think anything could run after these commands anyways.  So `burn` and `decompress` instructions are allowed to always be replayed.

### Other ideas
- Other suggestions involve always appending to the `cl_items` table and potentially doing gap management natively in the database.

### DIFF
--- a/init.sql
+++ b/init.sql
@@ -130,6 +130,7 @@ create table asset_grouping
     seq         bigint                      not null
 );
 -- Limit indexable grouping keys, meaning only create on specific keys, but index the ones we allow
+create unique index asset_grouping_asset_id on asset_grouping (asset_id);
 create index asset_grouping_key on asset_grouping (group_key, group_value);
 create index asset_grouping_value on asset_grouping (group_key, asset_id);
 
@@ -142,6 +143,7 @@ create table asset_authority
     authority bytea                       not null,
     seq       bigint                      not null
 );
+create unique index asset_authority_asset_id on asset_authority (asset_id);
 create index asset_authority_idx on asset_authority (asset_id, authority);
 
 -- creators
@@ -154,6 +156,6 @@ create table asset_creators
     verified bool                        not null default false,
     seq      bigint                      not null
 );
-
+create unique index asset_creators_asset_id on asset_creators (asset_id);
 create index asset_creator on asset_creators (asset_id, creator);
 create index asset_verified_creator on asset_creators (asset_id, verified);
--- a/init.sql
+++ b/init.sql
@@ -109,7 +109,8 @@ create table asset
     chain_data_id         bigint references asset_data (id),
     -- visibility
     created_at            timestamp with time zone     default (now() at time zone 'utc'),
-    burnt                 bool                not null default false
+    burnt                 bool                not null default false,
+    seq                   bigint              not null
 );
 
 create index asset_tree on asset (tree_id);
@@ -125,7 +126,8 @@ create table asset_grouping
     id          bigserial PRIMARY KEY,
     asset_id    bytea references asset (id) not null,
     group_key   text                        not null,
-    group_value text                        not null
+    group_value text                        not null,
+    seq         bigint                      not null
 );
 -- Limit indexable grouping keys, meaning only create on specific keys, but index the ones we allow
 create index asset_grouping_key on asset_grouping (group_key, group_value);
@@ -137,7 +139,8 @@ create table asset_authority
     id        bigserial PRIMARY KEY,
     asset_id  bytea references asset (id) not null,
     scopes    text[],
-    authority bytea                       not null
+    authority bytea                       not null,
+    seq       bigint                      not null
 );
 create index asset_authority_idx on asset_authority (asset_id, authority);
 
@@ -148,9 +151,9 @@ create table asset_creators
     asset_id bytea references asset (id) not null,
     creator  bytea                       not null,
     share    int                         not null default 0,
-    verified bool                        not null default false
+    verified bool                        not null default false,
+    seq      bigint                      not null
 );
-
 
 create index asset_creator on asset_creators (asset_id, creator);
 create index asset_verified_creator on asset_creators (asset_id, verified);
--- a/init.sql
+++ b/init.sql
@@ -96,6 +96,7 @@ create table asset
     supply_mint           bytea,
     -- compression
     compressed            bool                not null default false,
+    seq                   bigint              not null,
     -- -- Can this asset be compressed
     compressible          bool                not null default false,
     tree_id               bytea,
@@ -109,8 +110,7 @@ create table asset
     chain_data_id         bigint references asset_data (id),
     -- visibility
     created_at            timestamp with time zone     default (now() at time zone 'utc'),
-    burnt                 bool                not null default false,
-    seq                   bigint              not null
+    burnt                 bool                not null default false
 );
 
 create index asset_tree on asset (tree_id);
--- a/nft_ingester/src/main.rs
+++ b/nft_ingester/src/main.rs
@@ -182,7 +182,7 @@ async fn _handle_account(manager: &ProgramHandlerManager<'static>, data: Vec<(i6
 
 async fn handle_transaction(manager: &ProgramHandlerManager<'static>, data: Vec<(i64, &[u8])>) {
     for (message_id, data) in data {
-        println!("RECV");
+        println!("RECV, data size: {}", data.len());
         //TODO -> Dedupe the stream, the stream could have duplicates as a way of ensuring fault tolerance if one validator node goes down.
         //  Possible solution is dedup on the plerkle side but this doesnt follow our principle of getting messages out of the validator asd fast as possible.
         //  Consider a Messenger Implementation detail the deduping of whats in this stream so that
--- a/nft_ingester/src/main.rs
+++ b/nft_ingester/src/main.rs
@@ -182,7 +182,7 @@ async fn _handle_account(manager: &ProgramHandlerManager<'static>, data: Vec<(i6
 
 async fn handle_transaction(manager: &ProgramHandlerManager<'static>, data: Vec<(i64, &[u8])>) {
     for (message_id, data) in data {
-        println!("RECV, data size: {}", data.len());
+        println!("RECV");
         //TODO -> Dedupe the stream, the stream could have duplicates as a way of ensuring fault tolerance if one validator node goes down.
         //  Possible solution is dedup on the plerkle side but this doesnt follow our principle of getting messages out of the validator asd fast as possible.
         //  Consider a Messenger Implementation detail the deduping of whats in this stream so that
--- a/nft_ingester/src/parsers/bubblegum.rs
+++ b/nft_ingester/src/parsers/bubblegum.rs
@@ -18,6 +18,9 @@ use {
         SqlxPostgresConnector,
         TransactionTrait,
         DatabaseTransaction,
+        sea_query::OnConflict,
+        DbBackend,
+        ExecResult,
     },
     solana_sdk::pubkeys,
     digital_asset_types::{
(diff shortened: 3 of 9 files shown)


init.sql
```diff
@@ -109,7 +109,8 @@ create table asset
     chain_data_id         bigint references asset_data (id),
     -- visibility
     created_at            timestamp with time zone     default (now() at time zone 'utc'),
-    burnt                 bool                not null default false
+    burnt                 bool                not null default false,
+    seq                   bigint              not null
```
austbot, line 113: NIT i like to group the fields

austbot: so I wuld put this in the compression group this will be 0 or null for un compressed assets

nft_ingester/src/main.rs
```diff
@@ -182,7 +182,7 @@ async fn _handle_account(manager: &ProgramHandlerManager<'static>, data: Vec<(i6
 
 async fn handle_transaction(manager: &ProgramHandlerManager<'static>, data: Vec<(i64, &[u8])>) {
     for (message_id, data) in data {
-        println!("RECV");
+        println!("RECV, data size: {}", data.len());
```
austbot, line 185: pro debug

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -202,24 +229,32 @@ async fn handle_bubblegum_instruction<'a, 'b, 't>(
             let leaf_event = get_leaf_event(logs)?;
             db.transaction::<_, _, IngesterError>(|txn| {
                 Box::pin(async move {
-                    save_changelog_events(gummy_roll_events, slot, txn).await?;
+                    let seq = save_changelog_events(gummy_roll_events, slot, txn)
```
austbot, line 232: yeah this is a good Call I planned on this but we need a way to get a vec of these so we can guarantee multiple werent parsed, maybe I fixed this before with log msg splitting I think I did

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -232,21 +267,23 @@ async fn handle_bubblegum_instruction<'a, 'b, 't>(
...
                             let id_bytes = id.to_bytes().to_vec();
                             let asset_to_update = asset::ActiveModel {
                                 id: Unchanged(id_bytes.clone()),
                                 burnt: Set(true),
                                 ..Default::default()
                             };
-                            update_asset(txn, id_bytes, asset_to_update).await
+                            // Don't send sequence number with this update, because we will always
+                            // run this update even if it's from a backfill/replay.
+                            update_asset(txn, id_bytes, None, asset_to_update).await
```
austbot, line 286: why does burn not need seq?

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -369,17 +421,58 @@ async fn handle_bubblegum_instruction<'a, 'b, 't>(
...
+                            //             .to_owned(),
+                            //     )
+                            //     .build(DbBackend::Postgres);
+                            // query.sql =
+                            //     format!("{} WHERE excluded.seq >= asset.seq", query.sql);
+                            // txn.execute(query).await?;
+
+                            // Do not attempt to modify any existing values:
+                            // `ON CONFLICT ('id') DO NOTHING`.
+                            let query = asset::Entity::insert(model)
```
austbot, line 466: huh?

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -388,37 +481,121 @@ async fn handle_bubblegum_instruction<'a, 'b, 't>(
...
+                                //     .build(DbBackend::Postgres);
+                                // query.sql = format!(
+                                //     "{} WHERE excluded.seq >= asset_creators.seq",
+                                //     query.sql
+                                // );
+                                // txn.execute(query).await?;
+
+                                // Do not attempt to modify any existing values:
+                                // `ON CONFLICT ('asset_id') DO NOTHING`.
+                                let query = asset_creators::Entity::insert_many(creators)
```
austbot, line 511: we do need to modify

nft_ingester/src/parsers/bubblegum.rs
austbot: oh this is mint

nft_ingester/src/parsers/bubblegum.rs
austbot: aha in mint we dont I get it now

nft_ingester/src/parsers/bubblegum.rs
```diff
@@ -388,37 +481,121 @@ async fn handle_bubblegum_instruction<'a, 'b, 't>(
...
+
+                                // Insert into `asset_authority` table.
+                                let model = asset_authority::ActiveModel {
+                                    asset_id: Set(id.to_bytes().to_vec()),
+                                    authority: Set(update_authority),
+                                    seq: Set(seq as i64), // gummyroll seq
+                                    ..Default::default()
+                                };
+
+                                // Upsert way.
```
austbot, line 528: pet peev is commented out code in git

austbot (changes requested):
This is an interesting PR but it makes sense to me now. Can you clean up the commented out code since we have git we can always come back to it.

nft_ingester/src/parsers/bubblegum.rs
danenbm: yeah I will remove it, I just wanted to leave it in case we talked and we wanted it that way instead

nft_ingester/src/main.rs
danenbm: oh that was actually added since Sheldon had asked me about the data sizes, I can remove it

init.sql
danenbm: I'm not sure exactly what you mean, wondering if you are asking me to create some kind of group called "compression" but the only group concept I know of for SQL is user groups and `GROUP BY` statement.

Are you asking me to create another table or just make `burnt` able to be NULL since it will be NULL for uncompressed?

nft_ingester/src/parsers/bubblegum.rs
danenbm: I looked at `parse_logs` and I'm not seeing how it would prevent logs from multiple Bubblegum instructions from all being parsed into the same Vec.  I might be misunderstanding but seems like if somehow there were multiple bubblegum instructions in a single transaction, there could be logs from all of them in the transaction's logs.

Maybe this will become irrelevant if we start parsing candyguard instructions instead of using logs?

nft_ingester/src/parsers/bubblegum.rs
danenbm: It would be nice if we could use `seq` with every Bubblegum instruction.  However, in order to do this, every Bubblegum instruction would need to provide all the same fields.  As an example, `burnt` is not part of the on-chain leaf schema.  On-chain, the leaf is zeroed out, which is the same as with `redeem` except for the voucher.  Thus, if we missed a `burn` instruction and skipped writing the `burnt` flag due to `seq` number, we would not know if our off-chain asset was burned or redeemed.

So we need to actually always replay the `burn` instruction to make sure we set the `burnt` flag, and that only works because `burn` is orthogonal to the other instructions that modify `id`, `leaf`, `delegate`, and `owner`.

### Overall
In general, to make using a sequence number for the whole table work, there are two safe solutions:
1. Instructions must update all the same fields (`id`, `leaf`, `delegate`, `owner`).  Then if the table already has a greater `seq` number, we know the table already has newer data for those same fields (`id`, `leaf`, `delegate`, `owner`) so we don't rerun the instruction and risk putting in older stale data.
2. Instructions must only update unique fields.  `Burn` only updates a 'burnt' flag and nobody else does.  `Burn` also does not update `id`, `leaf`, `delegate`, or `owner`.  So `burn` can be rerun out of order and won't overwrite old data to anything (unless there was an `unburn` which doesn't make sense).

### Mint and decompress
You will note that both `mint` and `decompress` don't actually follow the two rules above, so then we have to just reason about them.

#### Minting
Any instructions after `mint` that change `id`, `leaf`, `delegate`, or `owner` will attempt to do an `UPDATE` to the existing `asset` table entry.  So they will actually fail if we haven't done the initial asset creation for the `mint` instruction processing.

Also, if the entry already exists and and we were to replay the `mint` without checking `seq`, we would potentially overwrite old data for `id`, `leaf`, `delegate`, or `owner`.

We could check `seq` and only rerun the `mint` instruction if we have a newer `seq` number, but that doesn't really make sense because we shouldn't ever be seeing a newer `mint` instruction for the same asset - there will only be one `mint` for an asset.

#### Decompress
After an asset is decompressed it is no longer managed by Bubblegum and Gummyroll so there can no longer be Bubblegum/Gummyroll events after this.

Also, `seq` is not part of the decompress event that is picked up by plerkle.  So yes this instruction modifies the `id` and `leaf` but it is currently a one-way door for the asset so if we missed it, we would rerun it because nothing will come after this.

Note this assumption falls apart if we add `compress` or other Bubblegum instructions that could occur after an asset was decompressed.

### Summary
I think what is in this PR will currently work with backfilling/replaying old transactions.  But it seems very fragile and not extensible, and it would be better to have some kind of versioning on every data field in the asset and related tables.

nft_ingester/src/parsers/bubblegum.rs
danenbm: (See other comment for why mint is doing this)

nft_ingester/src/parsers/bubblegum.rs
danenbm: (See other comment for why mint is doing this)

nft_ingester/src/main.rs
danenbm: Removed

nft_ingester/src/parsers/bubblegum.rs
danenbm: Done

init.sql
austbot: just move the seq to be under the --compression comment

nft_ingester/src/parsers/bubblegum.rs
austbot: gummy roll is what I mean not bummble gum if there were two actions on gummy roll

init.sql
danenbm: oh that is simpler request than i thought 😄

nft_ingester/src/parsers/bubblegum.rs
austbot: thanks for the explanation ,this makes alot more sense to my oatmeal powered brain

austbot (approved):

init.sql
danenbm: Done

nft_ingester/src/parsers/bubblegum.rs
danenbm: Ok I did not fully know what to do about this so for now I changed it to return a `Vec` containing all the seq numbers for all the events it encounters, and the Bubblegum  instruction parser gets the `max()` out of the `Vec`, which is assuming that there's only one set of logs per Bubblegum instruction.

danenbm:
Addressed last comments from @austbot, merged in latest changes from main and retested basic case with backfilling a couple missed sequence numbers and it still works.  Merging into main.

PR #162, opened 2022-07-26
Bubblegum, Gummyroll, Gumball, SS, Utils Public NPM packages

samwise2:
In this PR:
- [x] Convert `utils`, `gummyroll`, `bubblegum`, `gumball-machine`, `sugar-shack` into public NPM sdks!

https://www.npmjs.com/package/@sorend-solana/utils
https://www.npmjs.com/package/@sorend-solana/gummyroll
https://www.npmjs.com/package/@sorend-solana/bubblegum
https://www.npmjs.com/package/@sorend-solana/gumball-machine
https://www.npmjs.com/package/@sorend-solana/sugar-shack

The CLI (Gumball), tests and indexer all work using these packages. It's going to take a bit of work to get everyone up and running using them: everyone needs to `yarn link` each SDK, and then `yarn link @sorend-solana/...` within each dir that depends on it. Otherwise, the cli, the indexer and the tests will work with the SDKs at the npm registry.

To come in subsequent PRs:
- [ ] Upgrade to `spl-token` ^0.2.0 instead of 0.1.x
- [ ] Upgrade `Gummyroll` from manual SDK to `Solita`
- [ ] Review tsconfigs to see if they meet our needs
- [ ] Potentially add custom hook to solita.js to automatically pull the `Anchor<Program>` type for each contract into the SDK and export it (currently we still pull these types from `contracts/target/types...`
- [ ] Create a minified build more suitable for web apps (this will still work fine w website for now)
- [ ] Add docs to README for each SDK (once we want devs using them)


contracts/package.json
```diff
@@ -1,24 +1,5 @@
...
-    "express": "^4.18.1",
-    "pg-promise": "^10.11.1",
-    "react": "^18.1.0",
-    "sqlite": "^4.1.1",
-    "sqlite3": "^5.0.8",
-    "typescript-collections": "^1.3.3"
-  },
   "devDependencies": {
-    "@metaplex-foundation/rustbin": "^0.3.1",
-    "@metaplex-foundation/solita": "^0.10.0",
```
ngundotra, line 21: 🔥🔥🔥

contracts/sdk/gummyroll/src/types/index.ts
```diff
@@ -0,0 +1,762 @@
+// TODO(sorend): do we want to export the Anchor type in the SDK? We can modify all imports from contracts/target if so, but then we might need to custom hook into anchor build/test to keep this up to date.
```
ngundotra, line 1: Ideally we would export the anchor type in the SDK, but this is tricky because anchor moves quickly as well.

But ideally yes

contracts/sdk/indexer/smokeTest.ts
```diff
@@ -1,6 +1,7 @@
 import { Program, web3 } from "@project-serum/anchor";
 import fetch from "node-fetch";
-import { getMerkleRollAccountSize, Gummyroll } from "../gummyroll";
+import { Gummyroll } from "../../target/types/gummyroll";
```
ngundotra, line 3: Actually, we should start stripping all of these Anchor types from our typescript codebase.

They were only used before we had Solita.

contracts/sdk/utils/tsconfig.base.json
```diff
@@ -0,0 +1,20 @@
+{
+  "include": ["./index.ts"],
+  "compilerOptions": {
+    "sourceMap": true,
+    "declaration": true,
+    "declarationMap": true,
+    "allowSyntheticDefaultImports": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
ngundotra, line 9: does this include `downLevelIteration: true` ?

contracts/sdk/utils/index.ts
```diff
@@ -63,7 +67,7 @@ export function val(num: bignum): BN {
 /// Convert a string to a byte array, stored as an array of numbers
 export function strToByteArray(str: string, padTo?: number): number[] {
   let buf: Buffer = Buffer.from(
-    [...str].reduce((acc, c, ind) => acc.concat([str.charCodeAt(ind)]), [])
+    [...str].reduce((acc: number[], c, ind) => acc.concat([str.charCodeAt(ind)]), [])
```
ngundotra, line 70: 👍

ngundotra (changes requested):
This is a great stopping place; this is already a huge PR.

I'm in favor of merging this in once we have a standard script to do the `yarn link`-ing.

Can you add script that sets this up? Once that's done, we can use this in the demo website

contracts/sdk/indexer/smokeTest.ts
ngundotra: We will do this in another PR

contracts/sdk/gummyroll/src/types/index.ts
samwise2: let's take this out when we upgrade the Gummyroll SDK to Solita, some of Gummyroll's manual SDK methods depend on this and it's probably not worth refactoring since the next PR will overwrite the whole thing

contracts/sdk/utils/tsconfig.base.json
samwise2: I don't think so, but it looks like adding `downLevelIteration: true` is a good call so going to add that to all package.jsons. 

Also it might be worth putting all of the SDKs into a common subdir like `/public-sdks` so that they can all share a base tsconfig (BUT they would all still need their own base configs to include the correct files). This is a re-org that we can hold off on for now until we have more evidence that each SDK won't need more individual tsconfigs to the point where it wouldn't make sense

samwise2:
> This is a great stopping place; this is already a huge PR.
> 
> I'm in favor of merging this in once we have a standard script to do the `yarn link`-ing.
> 
> Can you add script that sets this up? Once that's done, we can start using this in the demo website

👍 added `candyland/link-local-sdks.sh` and `candyland/unlink-local-sdks` to automatically handle refreshing all node_modules and linking/unlinking all dependencies for `cli` and `contracts`, great idea!

PR #166, opened 2022-08-04
Renaming indexer to prototype-indexer

jnwng:
We all serve at DAS' pleasure


jarry-xiao (approved):

jnwng:
taken care of in #167

PR #171, opened 2022-08-11
CMT: Subtree Append

samwise2:
This PR adds functionality to append a subtree to a CMT!

The high-level process roughly goes as follows (similar to what @jarry-xiao initially outlined):
- Check that we are NOT trying to initialize the tree with an append (see below)
- Check that the subtree proof is valid for the subtree root
- Compute the new root of the tree by hashing the right most subtree leaf with other nodes in its rightmost proof, and then with the critical node (in the larger tree), and then with other nodes in the rightmost tree's RMP.
- Writeback the new changelog corresponding to the RML in the subtree
- Update the RMP index by incrementing it by the subtree's rightmost index

Why we only need one changelog in order to accurately fast-forward proofs for subsequent `replace_leaf` calls.
[Subtree append.pdf](https://github.com/jarry-xiao/candyland/files/9311928/Subtree.append.pdf)

Why we don't allow initializing a CMT via subtree append (for now, worth for the future?):
[Subtree append 2.pdf](https://github.com/jarry-xiao/candyland/files/9311930/Subtree.append.2.pdf)

### DIFF
--- a/lib/concurrent-merkle-tree/src/merkle_roll.rs
+++ b/lib/concurrent-merkle-tree/src/merkle_roll.rs
@@ -144,7 +144,7 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
         }
     }
 
-    /// Basic operation that always succeeds
+    /// Append leaf to tree
     pub fn append(&mut self, mut node: Node) -> Result<Node, CMTError> {
         check_bounds(MAX_DEPTH, MAX_BUFFER_SIZE);
         if node == EMPTY {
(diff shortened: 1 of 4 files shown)


lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -196,6 +196,91 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+    ) -> Result<Node, CMTError> {
+        check_bounds(MAX_DEPTH, MAX_BUFFER_SIZE);
+        // TODO: consider adding check that the subtree is not empty (but will omit for now)
+        //       note: it will be more time consuming to check that the subtree_root is not a root to an empty tree, because it requires applying O(depth) hashes of trivial nodes
+        if self.rightmost_proof.index >= 1 << MAX_DEPTH {
+            return Err(CMTError::TreeFull);
+        }
+
+        if self.rightmost_proof.index == 0 {
+            // TODO(sorend): Consider if we want to be able to initialize a tree with a subtree append (I think it would be cool, but is a little complex). write an equivallent function for subtree_append: initialize_tree_from_subtree_append
```
jarry-xiao, line 215: I think 0 should always be a safe index to start from because the subtree topology doesn't interfere with the main tree topology (i.e. each node can be mapped cleanly to another node)

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -196,6 +196,91 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+            }
+        }
+
+        self.update_internal_counters();
+        self.change_logs[self.active_index as usize] = ChangeLog::<MAX_DEPTH>::new(
+            node,
+            change_list,
+            self.rightmost_proof.index + subtree_rightmost_index - 1,
+        );
+        self.rightmost_proof.index = self.rightmost_proof.index + subtree_rightmost_index;
```
jarry-xiao, line 279: I think this can potentially mess up the tree

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -196,6 +196,91 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+                // Compute proof to the appended node from empty nodes
+                hash_to_parent(
+                    &mut intersection_node,
+                    &self.rightmost_proof.proof[i],
+                    ((self.rightmost_proof.index - 1) >> i) & 1 == 0,
+                );
+                hash_to_parent(
+                    &mut node,
+                    &subtree_rightmost_proof[i],
+                    ((subtree_rightmost_index - 1) >> i) & 1 == 0,
```
jarry-xiao, line 256: This logic seems like the right idea!

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -196,6 +196,91 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+        ) != subtree_root
+        {
+            return Err(CMTError::InvalidProof);
+        }
+
+        let leaf = subtree_rightmost_leaf.clone();
+        let intersection = self.rightmost_proof.index.trailing_zeros() as usize;
+
+        // @dev: At any given time (other than initialization), there is only one valid size of subtree that can be appended
+        if subtree_rightmost_proof.len() != intersection {
```
samwise2, line 234: @jarry-xiao here we enforce that we only append a tree of the correct size given the nodes inserted into the tree so far

lib/concurrent-merkle-tree/src/merkle_roll.rs
samwise2: agree, in general appending to an empty tree should be ok. I think the reason why I skipped it here for now was because there are multiple sizes of trees that are valid to append on initialization, and I was unsure if it would gel well with the logic for appending to an initialized tree. 

But it might be as simple as making a different size check on the depth of tree being appended if the tree is initially empty. will review the logic here

lib/concurrent-merkle-tree/src/merkle_roll.rs
jarry-xiao: Ah this is quite clever! Makes sense, because the intersection nodes maps to the root of the subtree where the newly appended node is the left most child

lib/concurrent-merkle-tree/src/merkle_roll.rs
jarry-xiao: Addressed due to the intersection check

jarry-xiao:
I think the logic here is looking sound after your explainer comment. The main thing that still seems somewhat uncertain to me is how to actually go about performing a tree migration. Operationally, it seems difficult to "wait" for the desired indices to become available for append. Maybe this can be handled purely from the controlling contract level where aggregated trees completely ban regular appends and only allow for subtrees of a certain size to be appended. Still, this is definitely a step in the right direction

lib/concurrent-merkle-tree/src/merkle_roll.rs
```diff
@@ -196,6 +196,91 @@ impl<const MAX_DEPTH: usize, const MAX_BUFFER_SIZE: usize> MerkleRoll<MAX_DEPTH,
...
+                );
+                hash_to_parent(
+                    &mut node,
+                    &subtree_rightmost_proof[i],
+                    ((subtree_rightmost_index - 1) >> i) & 1 == 0,
+                );
+                self.rightmost_proof.proof[i] = subtree_rightmost_proof[i];
+            } else if i == intersection {
+                // Compute where the new node intersects the main tree
+                hash_to_parent(&mut node, &intersection_node, false);
```
jarry-xiao, line 261: To check my understanding, at this point `node == root(subtree)`. For extra precaution, can we assert that here?

